PR #46, opened 2013-08-19
PNG fallback for old browsers that don’t support SVG

TomOne:
This is a pull request for a PNG fallback for old browsers that don’t support SVG (as announced at https://github.com/chocolatey/chocolatey/issues/324)). It’s a little JavaScript code which replaces all package icons in SVG format with their PNG equivalents that must be stored at the same location, for example http://tomone.github.io/icons/seafile.svg >>> http://tomone.github.io/icons/seafile.png.

In addition this makes the inline JavaScript code in the img tags (`onerror=…`) obsolete, because it already includes the fallback to the default package icon if an error occurs. Inline JavaScript event handlers are bad practice anyway.

The `pngfallback.js` script works fine. But I couldn’t figure out how to integrate this in the .cshtml files. It must be integrated at the bottom of the html body on every site which includes package icons.


ferventcoder:
Howdy, let's close this PR. There are some things it was looking to correct that it may have removed functionality for falling back to an image. I think we also won't need the fallback, although there may have been a resurgence in not loading SVG due to security reasons. However, if it's easy enough, we should be able to reimplement this functionality.

PR #116, opened 2014-11-02
Set UseDefaultCredentials to true on WebClient for web proxy authentication

patrickmaurer:
Setting this property to true (default is false unfortunately) will use the default credentials to authenticate with the web proxy in case one is present. For corporate users behind proxies this will enhance chances that chocolatey bootstrapping will work out of the box (copy from home page and run in cmd.exe or PowerShell). Hope I got the Razor markup right, checked it with a new MVC project in Visual Studio and it looked good, but wasn't able to run the real website solution for some reason.

### DIFF
--- a/chocolatey/Website/Views/Pages/Home.cshtml
+++ b/chocolatey/Website/Views/Pages/Home.cshtml
@@ -17,10 +17,10 @@
 </section>
 <div class="nuget-badge">
     <p>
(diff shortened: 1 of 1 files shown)


ferventcoder:
Would this be better suited to the advanced install options on the wiki?

patrickmaurer:
It depends. Don't know if many potential users of choco are behind a proxy that requires authentication. Also don't know, if first attempt to install fails (with command on home page), if these users will check the wiki for a solution.

ferventcoder:
I added this to the wiki, which drives this page now. https://github.com/chocolatey/choco/wiki/Installation#installing-behind-a-proxy

Which will become https://chocolatey.org/install#installing-behind-a-proxy

PR #141, opened 2015-01-03
(maint) Update Latest Release Number

gep13:
- Now that 0.9.8.29 has been released, let's update chocolatey.org to
reflect this
- Do we have a documented release process, so that this doesn't get
overlooked in future?

### DIFF
--- a/chocolatey/Website/Views/Pages/Home.cshtml
+++ b/chocolatey/Website/Views/Pages/Home.cshtml
@@ -37,7 +37,7 @@
 </section>
 
 <section class="release">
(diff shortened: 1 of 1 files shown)


ferventcoder:
:+1:

ferventcoder:
Nope, this is just a move by the seat of our pants kind of thing ha:)

gep13:
> Nope, this is just a move by the seat of our pants kind of thing ha:)

Ha ha, ok, I think this is something we need to capture then :smile_cat: For the day that you get hit by a bus!

PR #92, opened 2014-09-05
Advise maintainers to answer questions publicly on Disqus.

Redsandro:
Whenever someone contacts package maintainers with a question, include a link to the package comment section:

> User xyz sends the following message to the maintainers of Package 'abc'.
> 
> ...
> 
> Package Url: http://chocolatey.org/packages/abc
> 
> __Pro-tip:__ It is recommended to (also) post your answer to [the abc comment section](http://chocolatey.org/packages/abc#disqus) so that the information is publicly available. This can prevent you from receiving the same question from multiple users.
> 
> Package comments: [http://chocolatey.org/packages/abc#disqus](http://chocolatey.org/packages/abc#disqus)
> 
> ---
> To stop receiving contact emails as a maintainer of this package, sign in to the Chocolatey Gallery and change your email notification settings.

PR #145, opened 2015-01-05
Advise Maintainers to post to Disqus

gep13:
- This is a fixed up pull request for the work that @redsandro did in
GH-92
- A rebase was required that he wasn't comfortable doing, so this will
replace his PR

### DIFF
--- a/chocolatey/Website/Services/MessageService.cs
+++ b/chocolatey/Website/Services/MessageService.cs
@@ -118,6 +118,12 @@ Package Url: {6}
 {3}
 
 Package Url: {6}
+
+<b>Pro-tip:</b> It is recommended to (also) post your answer to [the {2} comment section](http://chocolatey.org/packages/{2}#disqus) so that the information is publicly available.
+This can prevent you from receiving the same question from multiple users.
+
+Package comments: [http://chocolatey.org/packages/{2}#disqus](http://chocolatey.org/packages/{2}#disqus)
+
 -----------------------------------------------
 <em style=""font-size: 0.8em;"">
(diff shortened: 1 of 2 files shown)


Redsandro:
Thank you. :+1:

ferventcoder:
Closes #92

Redsandro:
I see this is in. But the link is not working for me. It takes me to the package page, but not to the disqus anchor. Anyone else have this? Tested on Chromium 39 / Ubuntu. If not, nevermind.

---

If so, I don't know why this is. It is as if the anchor doesn't register in this particular website. Is it dynamically added to the body later? If so, it doesn't work.

One workaround would be to have `location.hash='disqus'` on the `onLoad` event, or after the content is added if it happens dynamically. Of course, anchor names only work when they change, so we would have to come from something arbitrary, e.g.

```
onLoad() {
  if (location.hash=='disqus') {
    location.hash='disqusanchor';
    location.hash='disqus';
  }
}
```

Just a thought. There are probably better fixes.

gep13:
@Redsandro can you provide an example of the generated URL?

Redsandro:
https://chocolatey.org/packages/chocolatey#disqus

gep13:
Ah, I see what you mean.  The URL doesn't work until the disqus object is loaded on the page, which doesn't happen until you scroll down to the bottom.  This will need some work.

ferventcoder:
The anchor is actually there. It's something with all anchors - not just this one

ferventcoder:
Although that link worked for me, I see Chrome go back and forth on working properly with anchors.

PR #146, opened 2015-01-05
Changed Comment Post Url

gep13:
- Changed default value of ForceSSL to be false (so that it will work
locally)
- Updated notes above Disqus thread to indicate new notifications
- Following discussion with Rob, built up comment Post Url in a way that
it will hopefully work locally, and on staging and production


gep13:
@ferventcoder ok, I "think" that this covers everything that we discussed.

Thoughts?

gep13:
@ferventcoder ok, I "think" we are good to go now.

Final email looks like the following:

image

ferventcoder:
That is beautiful.

gep13:
@ferventcoder when are you planning on pulling the trigger on this and pushing to chocolatey.org? Thanks!

ferventcoder:
Done.

Issue #172, opened 2015-04-09
Non-secure content on secure pages


DanAtkinson:
The following page types are loading unsecure content, thus raising security warnings in some browsers:

* Package search page.
 * [Example - 'commitmonitor' search results](https://chocolatey.org/packages?q=commitmonitor)
 * **Warning:** Mixed Content: The page at 'https://chocolatey.org/packages?q=commitmonitor' was loaded over a secure connection, but contains a form which targets an insecure endpoint 'http://feedburner.google.com/fb/a/mailverify'. This endpoint should be made available over a secure connection.
* Package page.
 * [Example - commitmonitor package](https://chocolatey.org/packages/commitmonitor).
 * **Information:** The non secure elements seem to be because Gravatar images are loaded over HTTP rather than HTTPS.
 * **Warning:** Mixed Content: The page at 'https://chocolatey.org/packages/commitmonitor' was loaded over HTTPS, but requested an insecure image 'http://www.gravatar.com/avatar/fdde6bfaf2716a26937e9f7dd0f5028f?s=32&r=g&d=retro'. This content should also be served over HTTPS.
 * **Warning:** Mixed Content: The page at 'https://chocolatey.org/packages/commitmonitor' was loaded over HTTPS, but requested an insecure image 'http://www.gravatar.com/avatar/031f04f9c9b0de39bfb2d381682f99d2?s=32&r=g&d=retro'. This content should also be served over HTTPS.
 * **Warning:** Mixed Content: The page at 'https://chocolatey.org/packages/commitmonitor' was loaded over HTTPS, but requested an insecure image 'http://www.gravatar.com/avatar/d8f76dc7a42d72dd4481fb3d74ce54dc?s=32&r=g&d=retro'. This content should also be served over HTTPS.

Kind regards, Dan

DanAtkinson:
Linked to #166 - although this issue contains a further bug relating to the feedburner form.

ferventcoder:
Yeah, the feedburner url may be fine just to switch to https

DanAtkinson:
That's cool. I submitted a pull request to fix this issue. The Gravatar issue appears to have a back-story however so I thought it best not to touch it unless I take the time to understand it a little more.

ferventcoder:
Not sure what the backstory is on the gravatar.

DanAtkinson:
[...]

PR #174, opened 2015-04-09
Ensure feedburner subscribe posts to a secure page

DanAtkinson:
Ensure feedburner subscribe posts to a secure page


DanAtkinson:
Partially fixes #172.

ferventcoder:
What are all of the other changes here?

DanAtkinson:
That's a good question! I manually edited the single file in the webpage so my guess is that it perhaps changed the whitespace to make it unified. Was it mixed before?

Doing a [spilt diff](https://github.com/chocolatey/chocolatey.org/pull/174/files?diff=split) doesn't identify any other change, but for all intents and purposes, the http -> https change was the only one made.

Given that the file changes aren't immediately clear, my recommendation would be to reject this pull request and make the change manually yourself. :)

ferventcoder:
This is a duplicate of #167

PR #176, opened 2015-04-10
oData Feed for Submitted Packages

### LINKED ISSUE #175: Provide oData Feed for Submitted Packages
In order to begin automating the moderation of submitted packages

gep13:
Provide ability to query, via oData, only from the packages which are
currently in a submitted state.

Without this change, automated package moderation will not be able to
get started.

Closed #175

### DIFF
--- a/chocolatey/Website/App_Start/Routes.cs
+++ b/chocolatey/Website/App_Start/Routes.cs
@@ -263,6 +263,11 @@ namespace NuGetGallery
                 typeof(V2CuratedFeed));
 
             routes.MapServiceRoute(
+                RouteName.V2ApiSubmittedFeed, 
+                "api/v2/submitted", 
+                typeof(V2SubmittedFeed));
+
+            routes.MapServiceRoute(
                 RouteName.V2ApiFeed,
                 "api/v2/",
                 typeof(V2Feed));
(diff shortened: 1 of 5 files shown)


gep13:
While testing this feed in nuget package explorer, I was getting no results returned, even though I had two packages in the submitted state, however, I think that this is due to some internals within nuget package explorer.  When I looked at the raw data coming back from the oData feed in Excel, I could correctly see the 2 packages:

image

I suspect that these properties:

image

were stopping them from appearing in nuget package explorer.

@ferventcoder let me know what you think of the above, and what else I need to do.

ferventcoder:
Is listed may also have something to do with it?

gep13:
Ah, yeah, could be.

chocolatey/Website/App_Start/Routes.cs
```diff
@@ -263,6 +263,11 @@ public static void RegisterRoutes(RouteCollection routes)
                 typeof(V2CuratedFeed));
 
             routes.MapServiceRoute(
+                RouteName.V2ApiSubmittedFeed, 
+                "api/v2/submitted-feed",
```
ferventcoder: just `api/v2/submitted` will do fine, don't you think?

gep13: Yip, I am easy :-)  I was going by the existing curated-feed.  I can change that up though.

gep13:
@ferventcoder I have fixed up the point that you made about this pull request

PR #194, opened 2015-06-13
(GH-101) Package Information Urls

ferventcoder:
Add information urls to Chocolatey.org.

This relates to #101.


ferventcoder:
All of that work and I determined it was the download path... :/

PR #221, opened 2015-09-17
update install.ps1

kkdd:
respect proxy environment variable


ferventcoder:
This has been completed in other ways.

PR #258, opened 2015-11-07
(GH-257) Host images locally

### LINKED ISSUE #257: Introduce image resizer to "host" images
That way there is no mixed content on the pages. http://imageresizing.net/

ferventcoder:
Closes #257


ferventcoder:
@gep13 can you take a look at this and try to determine why svg images won't host locally? I'm not quite sure but it might be something with my configuration. Going to an svg url directly wants to download the file.

PR #272, opened 2015-11-28
(GH-268) Add package validation bits

### LINKED ISSUE #268: Add validation bits
This is the API call to submit validation similar to how we do for verification.

- [x] add API for receiving package validation information.
- [x] Provide results information in the api feed
- [x] Create an auto-validation enumeration
- [x] Allow moderators to override package validation failures so that the verifier can run even if a package doesn't pass validation
- [x] Allow admins to set a package as exempt from verification (this is like driver installations, etc).

ferventcoder:
Add API and other bits to submit package validation. See GH-268 for more details of what this implements.

Closes #268


ferventcoder:
@gep13 thoughts?

gep13:
This looks good to me, nothing immediately jumping out as a problem.  Only one comment, which is more of a general one.

chocolatey/Website/ApiController.generated.cs
```diff
@@ -81,6 +81,13 @@ public partial class ApiController {
         }
 
         [NonAction]
+        [GeneratedCode("T4MVC", "2.0"), DebuggerNonUserCode]
```
gep13: Should this file even exist in source control?  i.e. since it is auto-generated?

ferventcoder: It's not auto generated for me. So yes. Plus it is required for compile. Also a yes. Same for me as asking if web service generated feed code should be in source control.

gep13: Ok, fair point. Just wanted to check :-D

PR #459, opened 2016-12-13
Corrected programdata to use the environmental variable

ajdelgado:
Using c:\programdata cause problems if you Windows disk is not c:\


ferventcoder:
Howdy @ajdelgado - I'm not sure I see what you've changed. The diff only shows changes with a write output message.

ajdelgado:
Yes, it's weird I didn't change that line but others below and github is not showing them. I'm going to close this and do another PR with the right changes.

PR #462, opened 2016-12-23
Fix docs typo (minor)

darthwalsh:
Its not It's.


ferventcoder:
Thank you. I'm not sure if we have a note in the documetation view folder but for fixes to this, you need to aubmit that to https://github.com/chocolatey/choco-wiki. HTH

ferventcoder:
I will work on adding a readme type file to the very top of this folder

darthwalsh:
Maybe you can delete this documentation folder? Less duplication means less mistakes :D

ferventcoder:
The docs here are the output of a process that converts the markdown to cshtml. Hopefully we can eradicate most of that, but this was the first pass to dot org becoming the one stop for docs, info, and the community packages.

Issue #173, opened 2015-04-09
Package marked as unlisted but has been approved


DanAtkinson:
Hi there!

I have a package ([commitmonitor](https://chocolatey.org/packages/commitmonitor)) which is live and should be listed but is set as unlisted on the package management screen.
![image](https://cloud.githubusercontent.com/assets/23440/7065335/77810b88-deae-11e4-9e95-68b86311bee1.png)

The following is a breakdown of the packages for commitmonitor.
![image](https://cloud.githubusercontent.com/assets/23440/7065320/646bac60-deae-11e4-9152-75bd67127067.png)

If I click the 'no' link on the second package, I'm taken to a page where I'm told that `Permanently deleting packages is not supported, but you can control how they are listed.`.

This is fine, but the package should still be registered as listed OR it should be both listed and unlisted since the package contains both.

Kind regards, Dan

gep13:
I am going to assume that you are referring to version 1.10.2 on CommitMonitor, which, if you click here:

https://chocolatey.org/packages/commitmonitor/1.10.2

You will see that it is actually marked as rejected.  As a result, you can't change the listed status of the package.

Is that the one that you are referring to?

DanAtkinson:
It's correct that one of the versions was intentionally rejected (the version number wasn't correct - it  missing the build number) but that shouldn't mean that the entire package should be set as unlisted in the package management page.

ferventcoder:
looks like a bug

ferventcoder:
Duplicate of #104.

Issue #104, opened 2014-10-23
Don't show packages with *some* versions unlisted as wholly unlisted


bc3tech:
On the Manage My Packages page, if you have unlisted _any_ version of a package, the entire package shows under the "Unlisted" area.
I have some packages that I've unlisted past versions because the package itself had an error in the configuration and I didn't want peoples' machines to be screwed up. Doing this, then going to the Manage My Packages area to see that the _whole_ package is unlisted (when the latest version is still listed and searchable) is really confusing.

ferventcoder:
#173 was marked as a duplicate, but has quite a bit of information and visuals related.

PR #464, opened 2017-01-01
(GH-104) Fix listed packages

Thilas:
Display a package as listed if it has at least one listed version, as suggested in #173.

I also changed how version is determined: instead of getting the max version (on a string...), I order by version (using SemanticVersion) descending. This means that max version is displayed but its related published date as well for instance.

### DIFF
--- a/chocolatey/Website/Controllers/UsersController.cs
+++ b/chocolatey/Website/Controllers/UsersController.cs
@@ -22,6 +22,7 @@ using System.Net.Mail;
 using System.Security.Principal;
 using System.Web.Mvc;
 using System.Web.UI;
+using NuGet;
 using NuGetGallery.MvcOverrides;
 using PoliteCaptcha;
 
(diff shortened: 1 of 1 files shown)


chocolatey/Website/Controllers/UsersController.cs
```diff
@@ -178,15 +179,17 @@ public virtual ActionResult Packages()
             var user = userService.FindByUsername(currentUser.Identity.Name);
             var packages = packageService.FindPackagesByOwner(user);
 
-            var published = from p in packages group p by p.PackageRegistration.Id;
+            var published = from p in packages
+                            orderby new SemanticVersion(p.Version) descending
+                            group p by p.PackageRegistration.Id;
```
ferventcoder: 👍

chocolatey/Website/Controllers/UsersController.cs
```diff
@@ -178,15 +179,17 @@ public virtual ActionResult Packages()
...
 
             var model = new ManagePackagesViewModel
             {
                 Packages = from pr in published
                            select new PackageViewModel(pr.First())
                            {
                                DownloadCount = pr.Sum(p => p.DownloadCount),
-                               Version = pr.Max(p => p.Version),
+                               Listed = pr.Any(p => p.Listed)
```
ferventcoder: Thank you

Issue #379, opened 2016-04-20
An asynchronous call is already in progress. It must be completed or canceled before you can call this method


ferventcoder:
"An asynchronous call is already in progress. It must be completed or canceled before you can call this method" was an error that came back when one of the services was attempting to update the site.
- http://stackoverflow.com/q/8369614/18475

Thilas:
Hey,
I just got this error while pushing 2 automatic packages (with AU from AppVeyor). Here is the report: https://gist.github.com/Thilas/c09573f2f23bc2aa904d90ff63de2cfe/787a9ea2fe2ca0f93bf362bb0a9a0f661ceb6a35. For the build on AppVeyor, it's this one: https://ci.appveyor.com/project/Thilas/chocolatey-packages/build/219.
Thanks,
Thomas

ferventcoder:
Thanks @Thilas - it's not often we see this. I see a possible fix in http://stackoverflow.com/a/16438292/18475

Thilas:
The idea seems nice indeed. I would prefer to do this using Task on my side. If you want I can submit a PR about it?

ferventcoder:
Sure. A PR would be lovely.

PR #476, opened 2017-02-08
(GH-379) Add sync lock on send mail to avoid async issues

Thilas:
I use a basic lock to make sure we don't try to simultaneously send mail on different threads and on the same SmtpClient instance.

Since emails need to be sent synchronously (according to what I saw in the code), it makes more sense to do fix this issue this way instead of using the ThreadPool (eg http://stackoverflow.com/a/16438292/18475) or Task (like I suggested).

Otherwise, an other solution would be to instantiate SmtpClient each time we need it instead of using the current singleton.

### DIFF
--- a/chocolatey/Website/Services/MailSender.cs
+++ b/chocolatey/Website/Services/MailSender.cs
@@ -20,6 +20,7 @@ using System;
 using System.IO;
 using System.Net.Mail;
 using System.Net.Mime;
+using System.Threading;
 using AnglicanGeek.MarkdownMailer;
 using MarkdownSharp;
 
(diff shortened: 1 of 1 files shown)


ferventcoder:
I'd like a Monitor with a timeout, otherwise this could be waiting for a long period of time. Thoughts?

Thilas:
Good idea. I set a timeout of 1 sec. Do you think it's ok?

ferventcoder:
Maybe give it about 5 seconds.

Thilas:
Done.

Thilas:
@ferventcoder, is there anything I can do on this PR in order to get it merged? Thanks.

Thilas:
Fixes #379.

@ferventcoder, I just noticed this PR was still open. Aren't you interested in merging it?

gep13:
@Thilas your changes have been merged, thanks for your contribution 👍

PR #484, opened 2017-03-12
typical requirement is PowerShell v3+

photodude:
typical requirement is PowerShell v3+
cmd.exe or PowerShell v2 etc are non-standard options in the more options section.


photodude:
How about we specify PowerShell v3+ as recommended and v2+ as minimum. 

It's a little confusing just saying v2+ as the requirement as the install instructions for v2+ are far more complicated than for v3+

ferventcoder:
@photodude it's still copy paste. ;) 

We are planning to move those v2 back up to the forefont on the install page. 

But also, if you noticed the readme in this folder, you would have seen that all of this comes from https://github.com/chocolatey/choco-wiki - that's where this will need to be filed.

ferventcoder:
"this folder" = the file at the very top of the documentation folder where you made the changes.

PR #494, opened 2017-04-08
(GH-491) Added check for 7za.exe already downloaded

### LINKED ISSUE #491: Feature request: only download 7za.exe if it is not already present
Would it be possible to add a 'Test-Path' statement to see if 7za.exe already exists in the temp folder, and if so skip the download?

This is helpful in scenarios where there is no internet, but we can use DSC to ensure the file is present prior to chocolatey being installed.

At present I work around this by using windows compression, using a chunk of DSC like this:

    Environment chocolateyDownloadUrl
    {
        Name  = "chocolateyDownloadUrl"
        Value = "https://chocolatey.myaddress.co.nz/install.ps1"
    }
    
    Environment chocolateyUseWindowsCompression
    {
        Name  = "chocolateyUseWindowsCompression"
        Value = "true"
    }
    
    cChocoInstaller installChoco
    {
        InstallDir             = "$env:PROGRAMDATA\chocolatey"
        ChocoInstallScriptUrl  = "https://chocolatey.myaddress.co.nz/install.ps1"
        DependsOn = @(  "[Environment]chocolateyDownloadUrl",
                        "[Environment]chocolateyUseWindowsCompression" )
    } 

@ferventcoder - I was slightly confused by your comments [here](https://github.com/chocolatey/choco/issues/1228) about the infrastructure management tools handling this?  If it's what I think you mean, I agree :), which is why the above makes sure the necessary environment variables are in place, and sources the install.ps1 from an internal repo.

AdmiringWorm:
Without this PR the 7za.exe file is always downloaded wether the 7za.exe file already exists in the temp directory or not.

fixes #491


ferventcoder:
Rebased and merged into master at https://github.com/chocolatey/chocolatey.org/commit/5f40343d430c644c5143e88c8adb2efa7f31af45. Thanks for the contributions!

PR #503, opened 2017-04-29
Bundling optimization

chad-ramos:
References #497 

@ferventcoder - Being you expressed a desire to work with Cassette, I went ahead and got it working with the current `style.css` file that you are serving in production. 

### Issue
As Cassette went through its build logic, it was running up against a class style that it appeared to be interpreting as UTF16 but had an expectation of UTF32.  Of which, was throwing an out-of-range error.  Long story short, there is a bug inside of Cassette. 

### Fix
 The offending style was `.terminal-chrome__close`.  Removing it has zero effect on the UI.  I am assuming the content type of `\c02715` produces an **X** value, of which you have replaced with `&times;`.  That being said, I removed it from `style.css` and verified the UI was unchanged. 

### Optimization
There were a few low-hanging-fruit things I did in the style sheet.  The commit notes should clearly outline them. 

### Thoughts
Personally, I dislike Cassette.  There are odd constructs enforced to manipulate your desired outcome. Of which to me, speaks to its age.  

* Forcing a full build to view changes seems like a nightmare for development.
* No guarantee you will not run into this bug again.
* Building bundles around none explicit configuration is not intuitive to desired expectations.

Yes, it works but I believe there are better options available today that align a bit more with how the web works.

### Results
| style.css  | Size |
| ------------- | ------------- |
| Current Production  | 62.6kb  |
| Pull Request | 46.9kb  | 

Every kb counts, :smile:

Issue #497, opened 2017-04-14
Why are we serving un-minified assets? 16kb is so key to my experience....


chad-ramos:
### What You Are Seeing?
My oh my, in this day in age how is it possible that you are serving un-minified css to the client.  What, do you think we have time to wait for this file to just....download?

![browser](https://cloud.githubusercontent.com/assets/3170200/25028565/0fad21ca-207b-11e7-890e-3124c10efb3e.png)

### What is Expected?
I only expect the best.  Clearly, someone missed the boat when it comes to optimization. Sheesh...I want my 16kb back? 
![min](https://cloud.githubusercontent.com/assets/3170200/25028607/4c8a27dc-207b-11e7-943b-5c508a07f92a.png)

### How Did You Get This To Happen? (Steps to Reproduce)
Yeah, I think I just "went to your site".

:smile: All in good fun.  Keep up the good work.  

If you are interested, I am more than willing to set a Gulp pipeline to handle this.

ferventcoder:
You got that thing to minify?!

ferventcoder:
Or should I say minify and bundle is where this went wrong

chad-ramos:
@ferventcoder So in all seriousness, I don't what compelled me to take a look at the assets this site was serving but regardless, I did notice your "style" sheet was un-minified.  Looks like there was an attempt to get the CSS assets optimized through the ASP.NET Bundler and that it might have been abandoned in favor of hard coding with a cache buster.  Is this correct?  Was the Bundler something you could not get working or decided to abandon? 

I can try and put a PR together to see if I can get that working for you or if you're interested, explore something like Gulp to handle it for you.

ferventcoder:
Not bundler, it was an older technology. Perhaps we should explore the included bundling

chad-ramos:
@ferventcoder - After finally getting around to exploring this, I have the following to say...

### Implementing the bundler by way of Microsoft.AspNet.Web.Optimization

I actually forgot that this bundling service does not actually have a native way to bust cache.  Meaning, to make this work we would...

- Bring in the NuGet package.
- Scaffold the necessary configuration
- Create a new static "Helper" class that wraps the native Optimization functionality in order to produce the necessary hanging query string `\style.cs?v=some-string` needed to bust the cache.

The challenge would be to decide on a way to produce and maintain the {some-string} portion of the equation. As of now, you are simply hard coding it. 

[...]

PR #504, opened 2017-05-01
(GH-497) Bug/bundling optimization

chad-ramos:
References Issue #497 
Failed PR at #503 

@ferventcoder - Being you expressed a desire to work with Cassette, I went ahead and got it working with the current `style.css` file that you are serving in production. 

### Issue
As Cassette went through its build logic, it was running up against a class style that it appeared to be interpreting as UTF16 but had an expectation of UTF32.  Of which, was throwing an out-of-range error.  Long story short, there is a bug inside of Cassette. 

### Fix
 The offending style was `.terminal-chrome__close`.  Removing it has zero effect on the UI.  I am assuming the content type of `\c02715` produces an **X** value, of which you have replaced with `&times;`.  That being said, I removed it from `style.css` and verified the UI was unchanged. 

### Optimization
There were a few low-hanging-fruit things I did in the style sheet.  The commit notes should clearly outline them. 

### Thoughts
Personally, I dislike Cassette.  There are odd constructs enforced to manipulate your desired outcome. Of which to me, speaks to its age.  

* Forcing a full build to view changes seems like a nightmare for development.
* No guarantee you will not run into this bug again.
* Building bundles around none explicit configuration is not intuitive to desired expectations.

Yes, it works but I believe there are better options available today that align a bit more with how the web works.

### Results
| style.css  | Size |
| ------------- | ------------- |
| Current Production  | 62.6kb  |
| Pull Request | 46.9kb  | 

Every kb counts, :smile:

### DIFF
--- a/chocolatey/Website/Content/style.css
+++ b/chocolatey/Website/Content/style.css
@@ -2523,10 +2523,6 @@ img  { max-width: 100%; }
   box-shadow: 5px 7px 50px rgba(0,0,0,.2);
 }
 
-.terminal-chrome__close {
-  content: '\c02715';
-}
-
 .terminal-chrome-top {
   position: absolute;
   left: 0;
--- a/chocolatey/Website/Content/style.css
+++ b/chocolatey/Website/Content/style.css
@@ -104,8 +104,8 @@ button {
 }
 @-webkit-keyframes fa-spin  {
   0%  {
-    -webkit-transform: rotate(0deg);
-    transform: rotate(0deg);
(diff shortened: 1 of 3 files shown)


ferventcoder:
Code has a small conflict - rebased and merged at https://github.com/chocolatey/chocolatey.org/commit/142f6572172199461ba51da3988de417b060aa2b - Thanks so much!!!

chad-ramos:
My pleasure.  It is the least I could do.  Wish I could help out more.

PR #510, opened 2017-05-09
Fix CommandNotFoundException on 7zip install check

int3rrupt:
Fix the exception thrown when the Install.ps1 script checks whether 7zip is already installed. Script is using C style syntax instead of proper Powershell syntax.


ferventcoder:
Thanks!

ferventcoder:
LF is bad in scripts for authenticode - cleaning that up.

ferventcoder:
Thanks for this. Rebased and merged into master at https://github.com/chocolatey/chocolatey.org/commit/ee46ffe7eea6172bd1935a5b66210999d359b28b

PR #513, opened 2017-05-16
Fix spelling

tomkirbygreen:


ferventcoder:
Hi Tom,
 I was coming away from that post noting that you may not have received the best advice on how to fix this. If you will notice in this folder there is a _README file. It explains this is actually built by the files at https://github.com/chocolatey/choco-wiki.

Now I can accept this contribution as is, but I'll also need to make it over there, as this will get overwritten on next wiki update.

ferventcoder:
Thanks for your contribution, but as it was a simple typo fix and typos annoy me once I know they are there, I went ahead and made the adjustments to the wiki repo and pulled that in here at 6abd7e1eb84d7a0ef6bf9eddb05beb9a9eac4af0

PR #538, opened 2017-12-09
The thanks page doesn't have a title

coolreader18:
Doesn't have to be exactly this, just bringing it to your attention

(I'm not sure that this is valid code, since I didn't test it in visual studio, but I'd assume it should work)

### DIFF
--- a/nugetgallery/Website/Views/Users/Thanks.cshtml
+++ b/nugetgallery/Website/Views/Users/Thanks.cshtml
@@ -1,4 +1,7 @@
 ﻿@using NuGetGallery;
+@{
+    ViewBag.Title = "Thanks!";
+}
 
 
 


AdmiringWorm:
I'll be honest and say that I don't use the upload page itself on the webpage.
But with that said, I think it would be better to say `Thank your for Uploading` (or something along those lines), instead of just saying thanks.

But that may just be me though.

coolreader18:
Oh, as far as I know, this just shows up before and after you confirm your account email.

AdmiringWorm:
yeah, I read the modified files again, and you are correct.
It seems it's the one you get after you have registered (I somehow misread the last line in the edited file)

ferventcoder:
Thanks!

PR #539, opened 2018-01-04
Disclaimers and cookies pop ups

derickbailey:
This pull request turns the notice at the top of hte /packages page into a full-screen pop up notification, requiring you to acknowledge it before you can do anything else on the site.

It also adds a cookie policy pop-up to the bottom of the site, which can also be closed.

in both cases, closing the pop up will remain "closed" in that browser instance for up to 100 years, by setting a cookie to prevent the page from displaying the pop up.

any time the cookie is not present, the pop ups will show up again


chocolatey/Website/Website.csproj
```diff
@@ -31,6 +31,7 @@
     <IISExpressAnonymousAuthentication />
     <IISExpressWindowsAuthentication />
     <IISExpressUseClassicPipelineMode />
+    <UseGlobalApplicationHostFile />
```
ferventcoder: I wonder where this snuck in from. No worries...

ferventcoder:
Did you want to squash this down or do you want me to handle that?

chocolatey/Website/Website.csproj
derickbailey: that's odd... maybe my specific version of visual studio introduced them for soem reason? i'll take these out and squash the commits.

derickbailey:
alright, i took out that strange line in teh website.csproj and squashed the commits for the PR into my master branch

ferventcoder:
I completely forgot to tell you about our super heavy caching strategies - any server side code checks won't work so most things like this would need to be done client side. I filed and implement #542, but it won't go back out until #543 gets addressed.

PR #546, opened 2018-01-16
Add Ansible to the text of the homepage.

huslage:
Ansible was missing from the list of devops tools.


ferventcoder:
So "like" wasn't enough there? ;)

It's fine to add this. I mean, Chocolatey integrates with pretty much anything, so we may need to switch over to a bulleted list at some point. 👍

PR #563, opened 2018-04-30
Update HowToHostFeed.cshtml

scott64bt:
Updated requirements for ProGet Server

### DIFF
--- a/chocolatey/Website/Views/Documentation/HowToHostFeed.cshtml
+++ b/chocolatey/Website/Views/Documentation/HowToHostFeed.cshtml
@@ -334,21 +334,21 @@ Some notable references:</p>
 <h4 id="proget">ProGet</h4>
 <ul>
 <li>Windows Server</li>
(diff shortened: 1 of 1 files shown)


gep13:
@scott64bt thank you for submitting this PR.  However, the documentation in this section of the site is actually generated from here: https://github.com/chocolatey/choco-wiki  There is a readme in the folder for the file that you changed, but sometimes it is missed.  Are you in a position to submit the change to that repo instead?

ferventcoder:
Do note that the data here was drawn from Inedo's site, then combined with our recommendations. We are not necessarily recommending the minimums here. 

For instance - there is no way we'll recommend SQL Server Express for production requirements, so please don't put that in there when you submit to the wiki repo.

gep13:
@scott64bt thank you.  I will go ahead and close off this PR in favour of this one: https://github.com/chocolatey/choco-wiki/pull/85

PR #566, opened 2018-05-07
(docs) update chocolateyInstall.ps1 URLs

simon04:

PR #567, opened 2018-05-07
(docs) update/fix age of chocolatey project

simon04:
"over 4 years ago" is an understatement for this project which started in 2011. :smile:


ferventcoder:
Hi Simon, same note on this as #566 - however I'm about to hit this section so I can catch this one with the other changes. 👍

PR #573, opened 2018-05-29
(maint) update copyright year

jochenkirstaetter:
Hi,

small maintenance issue regarding proper copyright information on the website (footer area in layout) and the PowerShell scripts.

Regards, JoKi


gep13:
Where we can use `DateTime.UtcNow.Year` I definitely think we should.

Where we can't, i.e. in the .ps1 files, would something like `2017 - Present` make more sense?  That way, we don't have to edit each year?

@ferventcoder thoughts?

gep13:
@jochenkirstaetter your changes have been merged, thanks for your contribution 👍

PR #592, opened 2018-08-11
Fix typo

ChrisMaddock:
This paragraph had the sentence: `The You would pass your token like this:`. I just removed it - I think the "Example Usage" title makes things clear enough.

By the way - this page was a sweet experience to first using the `choco push` api - everything there ready to copy and paste! 😄 

Edit: Screenshot removed, key regenerated 🤦‍♂️


gep13:
I think I tend to agree with you here.  Thanks for taking the time to make this change.

It will go out the next time we push out a version of the website.

PR #604, opened 2018-11-16
Removing unnecessary character

ber5ien:
Removing unnecessary character 


gep13:
@ber5ien thank you for taking the time to update this, however, I am not going to be able to merge this PR.  As mentioned in this readme:

https://github.com/chocolatey/chocolatey.org/blob/master/chocolatey/Website/Views/Documentation/_README.md

Direct changes to the cshtml files is not the correct place to make this change.  If this change is still required, could I please ask that you update the corresponding section in the original markdown file.

PR #605, opened 2018-11-26
Make selection easier

okonomiyaki3000:
- Set the `code` part of `.nuget-badge-small` to `user-select: all` so that clicking it automatically selects the entire `install` command. 
- Set the `span` part to `user-select: none` so that the prompt is not selectable. 

I suggest putting classes on those elements to make them easier to target with CSS but it's not so important so I'll leave it at this.

### DIFF
--- a/chocolatey/Website/Content/style.css
+++ b/chocolatey/Website/Content/style.css
@@ -4046,4 +4046,8 @@ section.package .moderator-side .waiting {
   padding: 15px;
   float:right;
   width: 450px;
+  user-select: all;
+}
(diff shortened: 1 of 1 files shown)


okonomiyaki3000:
I see you've used my excellent ideas in your redesign! Good work! No longer need this PR.

designxsteph:
Yes! I meant to reply back to this after the launch. I saw this open PR during the redesign and added this feature in. Thanks for the idea!

okonomiyaki3000:
I'm glad you used it, the new design looks great too!

Issue #638, opened 2019-01-18
Forms allowed to submit without checking required box


designxsteph:
### What You Are Seeing?
The Contact Us form and the Report Abuse form have a required checkbox that must be checked before submission, however right now the form is submitting even if the box is not checked.

### What is Expected?
If the user tries to submit the form before checking the required box, the form should show a validation error stating that the box must be checked.

### How Did You Get This To Happen? (Steps to Reproduce)

1. Fill out either the Contact Us form or the Report Abuse form on a package, but do not check the required box.
2. Submit the form.

PR #639, opened 2019-01-18
(GH-638) Add validation to required checkboxes

designxsteph:
Validation for required checkboxes has been added to the ContactUs and
ReportAbuse view models. Validation error boxes will now appear if the
user does not check the box, just as if it would if the name or email
field was not entered.

-Minor CSS adjustments to facilitate the addition of an error box if
not checked.


gep13:
@designxsteph I have just taken this for a spin, and I have some questions about this.

I tested on the latest version of the master branch, and when attempting to submit a form without clicking the checkbox, I get the following on the client side:

image

And the form isn't submitted.  Using the code in this branch, I no longer get the client side validation, however I do get the following returned:

image

Which is good, as that means that the server side validation of the checkbox is working as expected.

Do you know why the client side validation would have been lost as a result of this change?  Or was it not working for you previously?

gep13:
@designxsteph ok, digging further into this...

Looks like the client side validation isn't currently working on the live version of chocolatey.org, but it is when running locally.

So, I see these as two separate issues.  I think that we should validate that on the server side as well.

Would you mind raising another issue for investigating the lack of client side validation on chocolatey.org?

gep13:
@designxsteph I have merged this change here: https://github.com/chocolatey/chocolatey.org/commit/8d4bbcc12c55287fb673717bf42f485a06ef48b3

gep13:
@designxsteph thanks for adding this!

Issue #645, opened 2019-01-30
"Contact Site Admins" form will not submit


designxsteph:
### What You Are Seeing?
The "Contact Site Admins" form does not submit after filling out all required fields. The page will refresh, but not submit.

### What is Expected?
The form should submit after filling out all required fields and redirect back to the last viewed package page.

### How Did You Get This To Happen? (Steps to Reproduce)

1. Go to any package
2. Click on "Contact Site Admins" on the left navigation
3. Fill out all fields, and see that the form does not submit or redirect.

JourneyOver:
Glad this was not something on my end, as I was trying to send a "Contact Site Admins" form  a day or two ago about a package, and every time I pushed submit it would always just seem to refresh the page so I was always unsure on if it actually had ever sent or not.

So glad to find out it's just a bug.

ferventcoder:
This has just been merged and will be pushed up in a little while.

ferventcoder:
Building now...

PR #646, opened 2019-01-30
(GH-645) Add ContactAdmins view model

designxsteph:
The Report Abuse form and the Contact Admins form previously shared the
ReportAbuse view model. With the addition of the required check box to
the ReportAbuse view model in a previous commit, this made the Contact
Admins form unable to submit. The Contact Admins form was expecting a
required check box element, when there was none.

A new view model has been created specifically for the Contact Admins
form, and all appropriate files have been updated to reflect the change.


chocolatey/Website/ViewModels/ContactAdminsViewModel.cs
```diff
@@ -0,0 +1,50 @@
+﻿// Copyright 2011 - Present RealDimensions Software, LLC, the original
```
ferventcoder: Not something you need to do, but just noticed we do need to update our headers.

PR #654, opened 2019-03-28
Fix a warning about missing Tls 1.1/1.2 protocols

kasiagauza:
A warning could be issued, even if the protocols are present and a
correct .Net Framework is installed.

It's enough to trigger the warning just when SSL is not supported - and
now it's absent in PowerShell 6.1.x, so the warning is always there.

It's because the "try {}" block tests the whole following line:
"(...) 3072 -bor 768 -bor 192 -bor 48" and fails on "-bor 48" part.

What's changed is that now foreach loop tries to set a highest
encryption available by testing the protocols one by one, from SSL until
Tls 1.2. "catch {}" block is triggered when a specific protocol is
absent and the final "if" block checks the final selected encryption
(if it's Tls 1.2 or 1.1).

The solution works for PowerShell 2.x - 6.1.x.

### DIFF
--- a/chocolatey/Website/Install.ps1
+++ b/chocolatey/Website/Install.ps1
@@ -86,30 +86,33 @@ Fix-PowerShellOutputRedirectionBug
 # will typically produce a message for PowerShell v2 (just an info
 # message though)
 
-# Set SSL 3.0 (48), then TLS 1.0 (192), then TLS 1.1 (768), finally TLS 1.2 (3072)
 # Use integers because the enumeration values for TLS 1.2 and TLS 1.1 won't
 # exist in .NET 4.0, even though they are addressable if .NET 4.5+ is
 # installed (.NET 4.5 is an in-place upgrade).
 
-# Enumerate through protocols list; print not supported protocols as warnings
-(@{ 'Tls 1.2' = 3072; 'Tls 1.1' = 768; 'Tls 1.0' = 192; 'SSL 3.0' = 48 }).GetEnumerator() | Sort-Object -Property key | ForEach-Object {
+# Enumerate through protocols list and print warning when protocol is not supported
+[int32]$protocolsNumber = 0
+
+(@{ 'Tls 1.2' = 3072; 'Tls 1.1' = 768; 'Tls 1.0' = 192; 'SSL 3.0' = 48 }).GetEnumerator() | ForEach-Object {
   $temp = $_
   try {
     [System.Net.ServicePointManager]::SecurityProtocol = $_.Value
+    $protocolsNumber += [int32]$_.Value
   } catch {
     Write-Output "[Warning] $($temp.Key) protocol not supported."
   }
 }
 
-# Get the highest available encryption
-# Print warning if SecurityProtocol is not TLS 1.2 or TLS 1.1
-# Print the encryption
-$securityProtocol = [System.Net.ServicePointManager]::SecurityProtocol.toString().ToLower()
+# Protocol setter and getter is splitted to two lines to be compatible with PS 2.x
+[System.Net.ServicePointManager]::SecurityProtocol = $protocolsNumber
+$protocols = [System.Net.ServicePointManager]::SecurityProtocol
 
-if ($securityProtocol -ne 'tls11' -AND $securityProtocol -ne 'tls12') {
+# Print available protocols and warn if TLS 1.2 or TLS 1.1 are absent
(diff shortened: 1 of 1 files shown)


ferventcoder:
Hello @kasiagauza! Thanks for this contribution. I'm not quite following what this resolves. Can you please create an issue and then reference this commit to that issue? We'd like to have something to track to here. I'll also follow up here on what I'm seeing that is removed that should not be.

chocolatey/Website/Install.ps1
```diff
@@ -85,15 +85,31 @@ Fix-PowerShellOutputRedirectionBug
 # PowerShell will not set this by default (until maybe .NET 4.6.x). This
 # will typically produce a message for PowerShell v2 (just an info
 # message though)
-try {
-  # Set TLS 1.2 (3072), then TLS 1.1 (768), then TLS 1.0 (192), finally SSL 3.0 (48)
-  # Use integers because the enumeration values for TLS 1.2 and TLS 1.1 won't
-  # exist in .NET 4.0, even though they are addressable if .NET 4.5+ is
-  # installed (.NET 4.5 is an in-place upgrade).
-  [System.Net.ServicePointManager]::SecurityProtocol = 3072 -bor 768 -bor 192 -bor 48
```
ferventcoder: These are ORed together, which means it supports all of these. I see this has been removed below, and should be ORed back in.

chocolatey/Website/Install.ps1
```diff
@@ -85,15 +85,31 @@ Fix-PowerShellOutputRedirectionBug
...
+# Set SSL 3.0 (48), then TLS 1.0 (192), then TLS 1.1 (768), finally TLS 1.2 (3072)
+# Use integers because the enumeration values for TLS 1.2 and TLS 1.1 won't
+# exist in .NET 4.0, even though they are addressable if .NET 4.5+ is
+# installed (.NET 4.5 is an in-place upgrade).
+
+# Enumerate through protocols list; print not supported protocols as warnings
+(@{ 'Tls 1.2' = 3072; 'Tls 1.1' = 768; 'Tls 1.0' = 192; 'SSL 3.0' = 48 }).GetEnumerator() | Sort-Object -Property key | ForEach-Object {
+  $temp = $_
+  try {
+    [System.Net.ServicePointManager]::SecurityProtocol = $_.Value
```
ferventcoder:
```suggestion
    [System.Net.ServicePointManager]::SecurityProtocol = [System.Net.ServicePointManager]::SecurityProtocol -bor $_.Value
```

Not tested yet, just suggested to take a look at.

ferventcoder:
Apologies, the Not Implementing label was added and removed to test if GitHub fixed the reported issue with labels recently I was running into.

chocolatey/Website/Install.ps1
kasiagauza: I've changed the part with the newest commit. Basically here I just test if a given protocol is available, one by one and then sum the whole, like 3072+768+192 in _$protocolsNumber_ variable, and pass the final value to [System.Net.ServicePointManager]::SecurityProtocol = $protocolsNumber. That way I catch the absent protocol (SSL), issue a correct warning and avoid triggering a wrong one (_old .NET Framework installed_). It may be done in a cleaner way, though.

chocolatey/Website/Install.ps1
kasiagauza: I've created an issue #655 to explain the issue.

If we leave the whole '-bor' group together, then running the script in PowerShell 6.1.x will trigger a warning _Unable to set PowerShell to use TLS 1.2 and TLS 1.1 due to old .NET Framework installed (...)_, which is wrong, because TLS 1.x protocols are there, but the SSL is missing. The part _-bor 48_ raises an exception.

Either the warning message should be changed or a PS version check added, but I proposed a solution to check a list of protocols, one by one, and issue a warning when a specific one is missing. Then pass the final value to _SecurityProtocol_.

And then the warning: _Unable to set PowerShell to use TLS 1.2 and TLS 1.1 due to old .NET Framework installed (...)_ is shown exactly when it should.

ferventcoder:
With the changes that released today, the minimum version of TLS required to contact Chocolatey.org is TLS v1.2. With that it greatly simplifies the script and we needed to get that released and out the door following the changes. https://chocolatey.org/blog/remove-support-for-old-tls-versions

That unfortunately invalidated this request, so I'm going to go ahead and close it. I do appreciate the work on this though.

PR #656, opened 2019-04-09
(GH-613) Fix OutOfMemoryException during push

mkevenaar:
Relates to #613 

(GH-613) Fix OutOfMemoryException during push
By switching from a stream to a tempfile on the filesystem
the memory footprint of the application is reduced.
By doing so a switch to `OptimizedZipPackage` is required.
This is the same implementation the NuGet Gallery is using
when it ran into similar issues.

### DIFF
--- a/chocolatey/Website/Controllers/ApiController.cs
+++ b/chocolatey/Website/Controllers/ApiController.cs
@@ -127,7 +127,11 @@ namespace NuGetGallery
                 return new HttpStatusCodeWithBodyResult(HttpStatusCode.RequestEntityTooLarge,String.Format(CultureInfo.CurrentCulture, Strings.PackageTooLarge, MAX_ALLOWED_CONTENT_LENGTH / ONE_MB));
             }
 
-            var packageToPush = ReadPackageFromRequest();
+            // Tempfile to store the package from stream. 
+            // Based on https://github.com/NuGet/NuGetGallery/issues/3042
+            var temporaryFile = Path.GetTempFileName();
+
+            var packageToPush = ReadPackageFromRequest(temporaryFile);
 
             // Ensure that the user can push packages for this partialId.
             var packageRegistration = packageSvc.FindPackageRegistrationById(packageToPush.Id, useCache: false);
@@ -183,10 +187,25 @@ any moderation related failures.",
             try
             {
                 packageSvc.CreatePackage(packageToPush, user);
+
+                packageToPush = null;
+
+                try
+                {
+                    System.IO.File.Delete(temporaryFile);
+                }
+                catch (Exception ex)
+                {
+                    return new HttpStatusCodeWithBodyResult(HttpStatusCode.InternalServerError, "Could not remove temporary upload file: {0}", ex.Message);
+                }
             }
             catch (Exception ex)
             {
                 return new HttpStatusCodeWithBodyResult(HttpStatusCode.Conflict, string.Format("This package had an issue pushing: {0}", ex.Message));
+            }
+            finally
+            {
+                OptimizedZipPackage.PurgeCache();
             }
 
             return new HttpStatusCodeWithBodyResult(HttpStatusCode.Created, "Package has been pushed and will show up once moderated and approved.");
@@ -240,7 +259,7 @@ any moderation related failures.",
             filterContext.Result = new HttpStatusCodeWithBodyResult(HttpStatusCode.InternalServerError, exception.Message, request.IsLocal ? exception.StackTrace : exception.Message);
         }
 
-        protected internal virtual IPackage ReadPackageFromRequest()
+        protected internal virtual IPackage ReadPackageFromRequest(string temporaryFile)
         {
             Stream stream;
             if (Request.Files.Count > 0)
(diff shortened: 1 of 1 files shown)


chocolatey/Website/Controllers/ApiController.cs
```diff
@@ -240,17 +255,23 @@ protected override void OnException(ExceptionContext filterContext)
...
         }
 
-        protected internal virtual IPackage ReadPackageFromRequest()
+        protected internal virtual IPackage ReadPackageFromRequest(string temporaryFile)
         {
             Stream stream;
             if (Request.Files.Count > 0)
             {
                 // If we're using the newer API, the package stream is sent as a file.
+                // ReSharper disable once PossibleNullReferenceException
```
gep13: @mkevenaar we are going to want to refactor this, not simply ignore the warning 😄

mkevenaar:
Based on the changed done for https://github.com/NuGet/NuGetGallery/issues/3042

chocolatey/Website/Controllers/ApiController.cs
```diff
@@ -240,17 +255,23 @@ protected override void OnException(ExceptionContext filterContext)
...
                 stream = Request.Files[0].InputStream;
             }
             else stream = Request.InputStream;
 
-            return new ZipPackage(stream);
+            using (var temporaryFileStream = System.IO.File.Open(temporaryFile, FileMode.OpenOrCreate, FileAccess.ReadWrite))
```
gep13: Can you add a using statement at the top of the file for this?

gep13: Thinking about it, are other parts of the codebase using a FileSystem abstraction, or is it using System .IO directly?

chocolatey/Website/Controllers/ApiController.cs
mkevenaar: I have removed the comment.

chocolatey/Website/Controllers/ApiController.cs
mkevenaar: System.IO is already on top of the file. This was the only way I was able to get a working build.

gep13: Looks like there is another `File` associated with the base Controller class, so fully declaring it is necessary.

gep13:
@mkevenaar thanks again for making these changes.  These look good to me, but before merging and pushing this up to chocolatey.org, I think we should get a review from @ferventcoder as well.

chocolatey/Website/Controllers/ApiController.cs
```diff
@@ -183,11 +187,26 @@ private ActionResult CreatePackageInternal(string apiKey)
...
+                    return new HttpStatusCodeWithBodyResult(HttpStatusCode.InternalServerError, "Could not remove temporary upload file: {0}", ex.Message);
+                }
             }
             catch (Exception ex)
             {
                 return new HttpStatusCodeWithBodyResult(HttpStatusCode.Conflict, string.Format("This package had an issue pushing: {0}", ex.Message));
             }
+            finally
+            {
+                OptimizedZipPackage.PurgeCache();
```
ferventcoder: Q: Does anyone know what this actually does? We may need to do this in some other places.

chocolatey/Website/Controllers/ApiController.cs
ferventcoder: System.IO.File is correct here.
I tend to grab a bit more here with regard to shared access as part of opening a file.

chocolatey/Website/Controllers/ApiController.cs
```diff
@@ -250,7 +269,12 @@ protected internal virtual IPackage ReadPackageFromRequest()
             }
             else stream = Request.InputStream;
 
-            return new ZipPackage(stream);
+            using (var temporaryFileStream = System.IO.File.Open(temporaryFile, FileMode.OpenOrCreate, FileAccess.ReadWrite))
+            {
+                stream.CopyTo(temporaryFileStream);
+            }
+
+            return new OptimizedZipPackage(temporaryFile);
```
ferventcoder: Are we holding a lock on this file? I guess I'm thinking of TOCTOU - although it probably doesn't apply given this is on the server and not a client machine.

chocolatey/Website/Controllers/ApiController.cs
AdmiringWorm: Looking at the code for it, it looks like it is just cleaning up temporarily cached packages. Probably just to prevent the server to run out of space.

AdmiringWorm: The paths being cleaned are the ones being added when the `EnsurePackageFiles` method is called.

ferventcoder: Thanks for checking on that!

ferventcoder:
Thanks for this! I've just pushed this up to the community repo. Be aware that if things stop working entirely, it means we need to find a different folder.

ferventcoder:
That's an AppHarbor thing based on writeable folders, but I think the temp path is likely going to be fine. Just wanted to throw that awareness out there.

PR #662, opened 2019-05-15
(GH-661) URL of AWS S3 bucket changed.

### LINKED ISSUE #661: Amazon Deprecating Path-Style URL's Moving To Use Virtual-Hosted URL's
[Amazon are deprecating path-style S3 URL's](https://devclass.com/2019/05/13/amazon-s3-user-you-may-need-to-change-the-way-you-fill-your-bucket/) (https://s3.amazonaws.com/bucketname/folder/file.ext) from September 2020 and moving to the virtual-hosted style of URL's (https://bucketname.s3.amazonaws.com/folder/file.ext) instead.

There has been a [follow-up](https://aws.amazon.com/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/) to say that they will only be deprecating path-style URL's for buckets created **after* 30 September 2020:

```
Out with the Old
In response to feedback on the original deprecation plan that we announced last week, we are making an important change. Here’s the executive summary:

Original Plan – Support for the path-style model ends on September 30, 2020.

Revised Plan – Support for the path-style model continues for buckets created on or before September 30, 2020. Buckets created after that date must be referenced using the virtual-hosted model.
```

However this will potentially lead to two different ways of referencing S3 in the same codebase and across all of the Chocolatey codebase. As this is a small change I believe it should be made now in case the stay of execution for path-style URL's is reversed.

mkevenaar:
I have changed the URL used for the S3 bucket to the new format. The new format will
be required for new buckets starting from September 30, 2020. This format is working
for currently existing buckets as well.

Fixes #661


mkevenaar:
@ferventcoder @gep13 I think this small change could be implemented. I believe this does not change anything in the working of the website, but that's something you are best to test.

Old URL:
`https://s3.amazonaws.com/{bucketName}/{fileName}`

The new URL will be:
`https://{bucketName}.s3.amazonaws.com/{fileName}`

ferventcoder:
I believe they keep around the old method so for now we can probably close this. If we need to change it later we can revisit this. Thanks!

Issue #684, opened 2019-07-28
Clickjacking on website


HridayBhatia:
Bug report for clickjacking

ferventcoder:
@HridayBhatia this isn't the proper method to disclose security issues. I'm guessing you also sent an email recently after learning the proper method.

HridayBhatia:
I apologize because I didn't knew the proper channel and first point of contact so opened an issue at GitHub. From future I will refrain. Kindly consider this bug and timeline for remediation.
Also I do not remember sending you an email, as of I cannot find them on my sent box. I humbly request you to share that email or reply on it so I can manage to confirm that it shouldn't belong to someone else.

AdmiringWorm:
@HridayBhatia the email to send security related issues is mentioned here: https://chocolatey.org/security#report-issue

and it is: `security [at] chocolatey dot io`

ferventcoder:
This has been released and is live.

ferventcoder:
Special thanks to @HridayBhatia  and FKL (name redacted awaiting permission) for pointing this out!

PR #686, opened 2019-08-06
(GH-684) Add X-Frame-Options header for IIS

designxsteph:
A snippet has been added to Web.config to allow IIS to send the X-Frame-Options header. This X-Frame-Options header is used to avoid clickjacking attacks by ensuring that content on Chocolatey.org can not be embedded into other sites.

More information about this response header can be found at https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options


gep13:
Relates to #684

PR #690, opened 2019-09-11
upgrade if already installed

Jekotia:
Per issue chocolatey#164, choco will preform an upgrade if it is already installed.

I chose to test for chocolatey's presence by evaluating the output of `powershell choco -v`  with regex because I was having reliability issues when simply testing for the success of `powershell choco -v`.

### DIFF
--- a/chocolatey/Website/Install.ps1
+++ b/chocolatey/Website/Install.ps1
@@ -17,9 +17,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 # =====================================================================
-$ChocoVersionRegex = '^(\d+)\.(\d+)\.(\d+)$'
-$testchoco = (powershell choco -v)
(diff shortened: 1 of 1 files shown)


Jekotia:
> Just curious, any reason why not using `Get-Command choco.exe` wouldn't be enough to check for existence?

I did not know that was a thing, it never surfaced in my research somehow. I guess I'm still a bit too accustomed to the Unix world.

Change made, with `-ErrorAction Ignore` to prevent worrying red error output (when choco.exe is not found) for the user.

ferventcoder:
@Jekotia thanks so much, but this is in conflict with a change we'll be making for reasons I can't highlight yet. So unfortunately we won't be moving forward with this. It will make more sense once the other part is out.

Please use `choco upgrade chocolatey` as that is the way to upgrade Chocolatey. Thanks.

PR #609, opened 2018-12-01
Update animation to use "choco install docker-desktop"

StefanScherer:
After discussion in ahmetb/docker-chocolatey#49 and riezebosch/BoxstarterPackages#27 I've decided to move the choco package "docker" to "docker-cli".

Now the landing page at chocolatey.org must be updated to use something more meaningful than "choco install docker". So here is a PR for it to use the latest Docker Desktop instead with a "choco install docker-desktop"

PR #697, opened 2019-09-23
Use docker-desktop package in animation

StefanScherer:
Thanks for the new Chocolatey website. I really like the new UI.

One improvement I want to send you with this PR is that the packer `docker` is deprecated. That was originally the Docker CLI where nothing else existed yet on Windows. The most relevant package using Docker on a recent Windows 10 machine is the package `docker-desktop`. 

See also #609


ferventcoder:
@designxsteph I merged this in.

PR #701, opened 2019-09-23
(GH-699) Fix relevance in search results

### LINKED ISSUE #699: Search results are not returning relevant results
### What You Are Seeing?
When attempting to do a search from within the packages area, the results are being sorted by relevance instead of popularity. This is making the desired results not appear at the top.

### What is Expected?
Searches should be significant to what is being searched

Go to the packages area, search for "Chocolatey". See that the "Chocolatey" package does not even appear on the first page.

ferventcoder:
Id and Title need to be boosted quite a bit more in relevance for
search results. Add them to the package index entity with a higher
boost on id and title.

- Increase Id-Exact field to 4.5f from 2.5f in the index
- Add an Title-Exact field and give it a 4.0f boost in the index
- In the AND query (conjunction), boost it from 2.0f to 4.0f
- Boost exact Id query from 2.5f to 7.0f
- Add exact title query to search against Title-Exact with a 5.0f boost
- Add wildcard Title query to search against title.
- Remove the download count boosting on search results by relevance

Closes #699

### DIFF
--- a/chocolatey/Website/Infrastructure/Lucene/LuceneSearchService.cs
+++ b/chocolatey/Website/Infrastructure/Lucene/LuceneSearchService.cs
@@ -216,7 +216,7 @@ namespace NuGetGallery
 
             // All terms in the multi-term query appear in at least one of the fields.
             var conjuctionQuery = new BooleanQuery();
-            conjuctionQuery.Boost = 2.0f;
+            conjuctionQuery.Boost = 4.0f;
 
             // Some terms in the multi-term query appear in at least one of the fields.
             var disjunctionQuery = new BooleanQuery();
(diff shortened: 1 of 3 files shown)


chocolatey/Website/Infrastructure/Lucene/LuceneSearchService.cs
```diff
@@ -274,12 +278,12 @@ private static Query ParseQuery(SearchFilter searchFilter)
...
-                // If searching by relevance, boost scores by download count.
-                var downloadCountBooster = new FieldScoreQuery("DownloadCount", FieldScoreQuery.Type.INT);
-                return new CustomScoreQuery(combinedQuery, downloadCountBooster);
-            }
+            //if (searchFilter.SortProperty == SortProperty.Relevance)
+            //{
+            //    // If searching by relevance, boost scores by download count.
+            //    var downloadCountBooster = new FieldScoreQuery("DownloadCount", FieldScoreQuery.Type.INT);
+            //    return new CustomScoreQuery(combinedQuery, downloadCountBooster);
+            //}
```
ferventcoder: Leaving this commented out for now instead of removing completely. Pretty sure we'll be able to remove this (and also pretty sure that this is affecting search results).

chocolatey/Website/Infrastructure/Lucene/PackageIndexEntity.cs
```diff
@@ -78,10 +78,14 @@ public Document ToDocument()
 
             // title fields
             // If an element does not have a Title, fall back to Id, same as the website.
-            var workingTitle = String.IsNullOrEmpty(Package.Title)
+            var workingTitle = string.IsNullOrEmpty(Package.Title)
                                    ? Package.PackageRegistration.Id
                                    : Package.Title;
 
+            field = new Field("Title-Exact", workingTitle.ToLowerInvariant(), Field.Store.NO, Field.Index.NOT_ANALYZED);
```
ferventcoder: decided to lower case this, although most title items are left as it. Might want to change this later.

PR #715, opened 2019-10-11
(GH-683) Convert package page to Markdig/ Add text editor

### LINKED ISSUE #680: Backslash in code block
### What You Are Seeing?
Backslash in code block is interpreted.

### What is Expected?
Code block should be verbatim.

Look at https://chocolatey.org/packages/sshfs

Compare it with source and rendering on GitHub.

### LINKED ISSUE #683: Migrate from MarkDownSharp to Markdig
Would be a good idea to migrate to Markdig (or any other [CommonMark compliant implementation](https://github.com/commonmark/commonmark-spec/wiki/List-of-CommonMark-Implementations#c-1)). Even [MarkDownSharp](https://github.com/StackExchange/MarkdownSharp) itself recommends CommonMark.

designxsteph:
This PR updates the package page to use Markdig as opposed to MarkdownSharp. In addition, a new library "simplemde" has been implemented to add a wizzywig type markdown editor to the comments/review area which will help user write better comments and allow previewing before submitting.

Fixes #683 
Fixes #680


chocolatey/Website/Controllers/PackagesController.cs
```diff
@@ -66,6 +69,11 @@ protected virtual bool UseHttps()
             this.autoCuratedPackageCmd = autoCuratedPackageCmd;
             Configuration = configuration;
             this.searchSvc = searchService;
+
+            MarkdownPipeline = new MarkdownPipelineBuilder()
+                 .UseSoftlineBreakAsHardlineBreak()
+                 .UseAdvancedExtensions()
```
AdmiringWorm: You may perhaps want to change this, as the extension `UseAdvancedExtension` may be pulling in some things that probably shouldn't be added, or isn't wanted (For instance: Extra emphasis, figures, grid tables, mathematics, media links, pipe tables, task lists, diagrams and generic attributes. Could be more)

reference: https://github.com/lunet-io/markdig/blob/master/src/Markdig/MarkdownExtensions.cs#L77-L94

ferventcoder: We like the advanced extensions for the most part. I think we may want to get more specific for the package pages on what we want to use and at least have what we already had with MarkdownSharp.

chocolatey/Website/Controllers/PackagesController.cs
```diff
@@ -168,6 +176,11 @@ public virtual ActionResult DisplayPackage(string id, string version)
 
             var scanResults = packageSvc.GetPackageScanResults(id, package.Version);
             package.Files = packageSvc.GetPackageFiles(package).ToList();
+            package.ReviewComments = Markdown.ToHtml(package.ReviewComments ?? string.Empty, MarkdownPipeline);
+            package.Description = Markdown.ToHtml(package.Description ?? string.Empty, MarkdownPipeline);
+            package.ReleaseNotes = Markdown.ToHtml(package.ReleaseNotes ?? string.Empty, MarkdownPipeline);
+            package.PackageRegistration.ExemptedFromVerificationReason = Markdown.ToHtml(package.PackageRegistration.ExemptedFromVerificationReason ?? string.Empty, MarkdownPipeline);
+
```
ferventcoder: Definitely would want to do something with the StringExtensions.clean_html() on these items.

ferventcoder: https://github.com/chocolatey/chocolatey.org/blob/0e11347b669c221c6a3364744f0b3b079e4c6556/chocolatey/Website/StringExtensions.cs#L81-L86

chocolatey/Website/Controllers/PackagesController.cs
ferventcoder: https://github.com/lunet-io/markdig/blob/f8ab1cccc5dc58626f8b5974c90d73067749d261/src/Markdig/MarkdownExtensions.cs#L74-L95

~~~
return pipeline
                .UseAbbreviations()
                .UseAutoIdentifiers()
                .UseCitations()
                .UseCustomContainers()
                .UseDefinitionLists()
                .UseEmphasisExtras()
                .UseFigures()
                .UseFooters()
                .UseFootnotes()
                .UseGridTables()
                .UseMathematics()
                .UseMediaLinks()
                .UsePipeTables()
                .UseListExtras()
                .UseTaskLists()
                .UseDiagrams()
                .UseAutoLinks()
                .UseGenericAttributes(); // Must be last as it is one parser that is modifying other parsers
~~~

ferventcoder: * `UseSoftlineBreakAsHardlineBreak()`
* `UseAutoLinks()`
* `UseGridTables()`
* `UseAutoIdentifiers()`
* `UseEmphasisExtras()`
* `UseNoFollowLinks()`

AdmiringWorm: You probably also would want to to have `DisableHtml()` as well, to prevent users including raw html without having to rely on the custom `clean_html` method (No harm in having both though, I think).

As well as using `UseCustomContainers()`, if this isn't used then using code blocks for specific languages wouldn't be possible.

ferventcoder: yeah, good point. I'll note this on the current commits.

chocolatey/Website/Controllers/PackagesController.cs
```diff
@@ -66,6 +69,15 @@ protected virtual bool UseHttps()
...
             Configuration = configuration;
             this.searchSvc = searchService;
+
+            MarkdownPipeline = new MarkdownPipelineBuilder()
+                 .UseSoftlineBreakAsHardlineBreak()
+                 .UseAutoLinks()
+                 .UseGridTables()
+                 .UseAutoIdentifiers()
+                 .UseEmphasisExtras()
+                 .UseNoFollowLinks()
```
ferventcoder: @AdmiringWorm [made a good point earlier]() about `UseCustomContainers(https://github.com/chocolatey/chocolatey.org/pull/715#discussion_r335537194)` and `DisableHtml()` - we'll want to find a good package to try that disable html on, but I think it does basically the same thing as `clean_html()` does.

chocolatey/Website/Views/Packages/_ListVersion.cshtml
```diff
@@ -1,12 +1,21 @@
...
     var maintainer = User != null && Model.Package.Owners.Any(u => u.Username == User.Identity.Name);
     var moderator = User != null && User.IsModerator() && !maintainer;
+    
+    var MarkdownPipeline = new MarkdownPipelineBuilder()
+                 .UseSoftlineBreakAsHardlineBreak()
+                 .UseAutoLinks()
+                 .UseGridTables()
+                 .UseAutoIdentifiers()
+                 .UseEmphasisExtras()
+                 .UseNoFollowLinks()
```
ferventcoder: That change would need to be applied here as well to add those two additional components.

ferventcoder: Actually, let's leave out the disable html as I read a bit more and I don't think we want to do that.

chocolatey/Website/Controllers/PackagesController.cs
ferventcoder: Actually, let's leave out the disable html as I read a bit more and I don't think we want to do that.

ferventcoder: So let's just add `UseCustomContainers()` to this list.

designxsteph:
Running into a bug with this now, and narrowed it down to usage of the clean_html(). Will update again when this is ready.

chocolatey/Website/Views/Packages/VerifyPackage.cshtml
```diff
@@ -30,7 +46,7 @@
                         <p>@Model.Title</p>
                     }
                     <h3 class="mb-0">Description:</h3>
-                    <p>@Model.Description</p>
+                    @Html.Raw(Markdown.ToHtml(Model.Description.clean_html() ?? string.Empty, MarkdownPipeline))
```
ferventcoder: Noted the `<p>` was removed from here. Was this on purpose?

designxsteph: This was done on purpose. Since the description itself is converted to html, it does not need to be wrapped in another `<p>` element.

ferventcoder: Cool!

PR #716, opened 2019-10-15
Global search

designxsteph:
In this PR:
* Adds a globals search box to the top navigation while on the home page
* Makes the search box inside the packages area more prominent
* Adds a search box in the documentation area


ferventcoder:
Oh and add a drop down with hints similar to stackoverflow search box.

id:searchValue and tag:searchValue

PR #726, opened 2019-10-28
Reduce need for JS

designxsteph:
In this PR:
* Updates preloader to use pure css animations so it is not reliant on javascript to render
* Reduces javascript used in files area on the package page by using razor syntax instead

### DIFF
--- a/chocolatey/Website/Views/Shared/_Loader.cshtml
+++ b/chocolatey/Website/Views/Shared/_Loader.cshtml
@@ -1,0 +1,6 @@
+﻿<div id="loader">
+    <div>
+        <img src="https://chocolatey.org/content/images/logo_small.svg" />
+        <div></div>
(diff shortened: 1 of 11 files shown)


ferventcoder:
This now has conflicts, I would like to see this up on choco test as well once we get through the others.

ferventcoder:
Is this fixing any issues (like earlier versions of IE)?

chocolatey/Website/Views/Shared/_Loader.cshtml
```diff
@@ -0,0 +1,6 @@
+﻿<div id="loader">
+    <div>
+        <img src="https://chocolatey.org/content/images/logo_small.svg" />
```
ferventcoder: @pauby @gep13 you folks will be so happy!

PR #733, opened 2019-11-14
(GH-732) Fix broken doc links

### LINKED ISSUE #732: Doc Urls broken from conversion to Markdig
With the conversion of the documentation area to use Markdig, a few special type links were broken resulting in 404s.

### 1. Links that contain both `-` and camel casing result in a 404
Currently there is one file ( How-To-Parse-PackageParameters-Argument.md ) that combines using `-` and camel casing in the name. There is no logic to support this right now which makes the page desired go to a 404.

Example: The correct link of https://chocolatey.org/docs/how-to-parse-package-parameters-argument results in a 404. Right now it is being parsed as https://chocolatey.org/docs/how-to-parse-packageparameters-argument (which works but is not the correct url)

### 2. One off links that don't fit a pattern
Right now there is one snowflake file `CommandsApiKey.md` that is being parsed correctly based on the camel case logic, but the URL it results to https://chocolatey.org/docs/commands-api-key is not the URL used in the past. The URL it should be going to is https://chocolatey.org/docs/commands-apikey. As a result, there needs to be some special logic to handle this one off file.

designxsteph:
This PR addresses the issues brought up in #732 which pertains to broken documentation links.

Fixes #732

### DIFF
--- a/chocolatey/Website/Controllers/DocumentationController.cs
+++ b/chocolatey/Website/Controllers/DocumentationController.cs
@@ -180,7 +180,13 @@ namespace NuGetGallery.Controllers
             Char previousChar = '^';
             foreach (var valueChar in fileName)
             {
-                if (Char.IsUpper(valueChar) && hyphenatedValue.Length != 0 && !Char.IsUpper(previousChar))
+                // Filenames that contain both a "-" and camel casing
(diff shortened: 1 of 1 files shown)


gep13:
@designxsteph I have just pushed this to production, and the broken link now seems to be working correctly.  Thanks for fixing this so quickly!

PR #744, opened 2019-11-27
Moderation Filtering and Search Result Fixes

### LINKED ISSUE #731: Moderator queue search results incorrect
### What You Are Seeing?
image

### What is Expected?
"Returned X packages" should match the verbiage under "Display results N - M of X" and 'Next' link at bottom should be enabled/disabled (shown/hidden) accordingly.

Filter the moderator queue w/ a package name as the `q` parameter in the url. e.g. https://chocolatey.org/packages?q=vivaldi&moderatorQueue=true&moderationStatus=all-statuses&prerelease=false&sortOrder=package-download-count

### LINKED ISSUE #741: Moderation status filter does not work when using search term
### What You Are Seeing?
While in the moderation queue if searching for a search term, the moderation status filter does not do anything. The filter can be changed, but no matter what is selected the results remain the same.

ModerationFilter

### What is Expected?
I would expect the moderation status to still work if searching for a term, or not be visible if not wanted in a search.

1. Search for "chocolatey"
2. Go to the moderation queue
3. Try to switch the moderation status filter to anything other than "All Moderation Statuses"

### LINKED ISSUE #743: Moderation status filtered pages do not reflect current count
### What You Are Seeing?
When going to the moderation queue, you will see the counts of "Updated, Responded, Submitted, Waiting, Unreviewed" packages. However, if you click on one of those to take you to a page with those filters applied, the count at the top does not reflect the same number shown on the first page. The correct number of packages is shown however.

mod2

When the "Submitted" is clicked on:
mod1

### What is Expected?
I would expect the top number on the clickthrough page to reflect the same number as on the overview area, and the same number of packages that is actually represented.

1. Go to the moderation queue
2. Click on "Submitted Packages" under the main moderation count heading. Or you could also use the moderation status dropdown filter.
3. Notice the numbers do not match as shown in the pictures above.

designxsteph:
This PR addresses 3 main issues that relate to package status filtering and incorrect number counts while in the moderation queue. 

Fixes #743 
Fixes #731 
Fixes #741

### DIFF
--- a/chocolatey/Website/ViewModels/PackageListViewModel.cs
+++ b/chocolatey/Website/ViewModels/PackageListViewModel.cs
@@ -39,6 +39,8 @@ namespace NuGetGallery
             int submittedCount,
             int waitingCount,
             int respondedCount,
+            int pendingAutoReviewCount,
+            int unknownCount,
             string moderationStatus)
         {
             // TODO: Implement actual sorting
@@ -82,6 +84,8 @@ namespace NuGetGallery
             ModerationSubmittedPackageCount = submittedCount;
             ModerationWaitingPackageCount = waitingCount;
             ModerationRespondedPackageCount = respondedCount;
+            ModerationPendingAutoReviewPackageCount = pendingAutoReviewCount;
+            ModerationUnknownPackageCount = unknownCount;
         }
 
         public int FirstResultIndex { get; set; }
@@ -114,5 +118,7 @@ namespace NuGetGallery
         public int ModerationSubmittedPackageCount { get; private set; }
         public int ModerationWaitingPackageCount { get; private set; }
         public int ModerationRespondedPackageCount { get; private set; }
+        public int ModerationPendingAutoReviewPackageCount { get; private set; }
+        public int ModerationUnknownPackageCount { get; private set; }
     }
 }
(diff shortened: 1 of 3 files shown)


chocolatey/Website/ViewModels/PackageListViewModel.cs
```diff
@@ -39,6 +39,9 @@ public class PackageListViewModel
             int submittedCount,
             int waitingCount,
             int respondedCount,
+            int unreviewedCount,
+            int pendingCount,
```
ferventcoder: call this out - `pendingAutoReviewCount`

chocolatey/Website/ViewModels/PackageListViewModel.cs
```diff
@@ -39,6 +39,9 @@ public class PackageListViewModel
             int submittedCount,
             int waitingCount,
             int respondedCount,
+            int unreviewedCount,
```
ferventcoder: This is being passed in as `readyPackagesCount`, but `unreviewedCount` here. The confusion is that there is an `unreviewedPackagesCount` also being passed to `submittedCount`, which by looking at it alone makes it look like we have a bug, but it's correct.

I think because we have pending auto review count, we can remove it from the unreviewed (aka ready) and remove this one entirely.

gep13:

Issue #754, opened 2019-12-02
Convert Search Filter index all versions


ferventcoder:
Currently requests for all versions always query against the cache/database, but it would be better to simply use the index instead.

PR #755, opened 2019-12-02
(GH-754) Index All Versions

ferventcoder:
Ensure all package versions are stored in the index to reduce calls to the database.

### DIFF
--- a/chocolatey/Website/Infrastructure/Lucene/LuceneIndexingService.cs
+++ b/chocolatey/Website/Infrastructure/Lucene/LuceneIndexingService.cs
@@ -142,13 +142,15 @@ namespace NuGetGallery
                 // We need to do this because some attributes that we index such as DownloadCount are values in the PackageRegistration table that may
                 // update independent of the package.
                 set = set.Where(
-                    p => (p.IsLatest || p.IsLatestStable) &&
+                    p => (_indexContainsAllVersions || p.IsLatest || p.IsLatestStable) &&
                          p.PackageRegistration.Packages.Any(p2 => p2.LastUpdated > lastIndexTime));
             }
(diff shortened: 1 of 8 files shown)


ferventcoder:
@vexx32 @gep13 if you can take a quick peek and give some feedback, I'll be back in a little over an hour to take a closer look at this.

chocolatey/Website/Infrastructure/Lucene/LuceneIndexingService.cs
```diff
@@ -140,14 +142,16 @@ private List<PackageIndexEntity> GetPackages(DateTime? lastIndexTime)
                 // We need to do this because some attributes that we index such as DownloadCount are values in the PackageRegistration table that may
                 // update independent of the package.
                 set = set.Where(
-                    p => (p.IsLatest || p.IsLatestStable) &&
+                    p => (_indexContainsAllVersions || p.IsLatest || p.IsLatestStable) &&
```
vexx32: You may want to update the above comment here. 🙂

ferventcoder: The comment looks to still apply? 

~~~csharp
// We need to do this because some attributes that we index such as DownloadCount are values in the PackageRegistration table that may
// update independent of the package.
~~~

This bit is about noting and capturing metadata related to a package that updates independently of the package itself, such as download counts. Previously, it was only making sure those things were updated with the Latest or Latest Stable versions, but now since we would be storing all versions, it would need to update any package that has had updates - older versions can have download counts adjust.

Does that make sense? Or were you saying there should be something different from what I'm saying noted in the comment?

PR #759, opened 2019-12-04
Configure Tags as NOT_ANALYZED

slang25:
When clicking on a tag from the package list, you are shown the search results for a query on tag with the term of the tag you clicked on. However, because tags are indexed with analysis on, any tags with a hyphen in will be indexed as multiple words.

This is the result:
https://twitter.com/sqldbawithbeard/status/1202348325808361474

This change would result in these becoming exact matches.

### DIFF
--- a/nugetgallery/Website/Infrastructure/Lucene/LuceneIndexingService.cs
+++ b/nugetgallery/Website/Infrastructure/Lucene/LuceneIndexingService.cs
@@ -133,7 +133,7 @@ namespace NuGetGallery
 
             if (!String.IsNullOrEmpty(package.Tags))
             {
-                field = new Field("Tags", package.Tags, Field.Store.NO, Field.Index.ANALYZED);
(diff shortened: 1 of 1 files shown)


ferventcoder:
Howdy - I've done a lot of work in this area recently that I have yet to push up, but I think this makes sense.

ferventcoder:
@slang25 unfortunately you hit a file we are not using - that is in the nugetgallery side. You might travel down chocolatey side to PackageIndexEntity.cs

ferventcoder:
I was working on another issue here, that when you search tags, it wasn't searching all items as tags. It might resolve this issue as well.

slang25:
Ah, I see the error of my ways. Those changes look good, but I don't think it will fix the issue (we shall see). Can we change `Tag` to `NOT_ANALYZED`, or are we better off adding a `Tag-Exact` thing?

ferventcoder:
I haven't forgotten about this, it just hasn't been implemented yet as it requires some more testing around the effects of this. 👍

PR #763, opened 2019-12-06
(GH-762)(GH-761) Lucene Index Fixes and Enhancements

### LINKED ISSUE #761: Updating a package causes multiple results
When using the lucene index for all results, if you update a package or push a newer version, you will notice multiple results on the community repository. 

image

That's because one got added with IsLatest, and the other one was still there with IsLatest.

### LINKED ISSUE #762: Search Filter Index Enhancements
Following on #754, there are some additional adjustments to make with respect to ensuring the cache is correct.

* Set timeout for connection itself to allow for more than 110 seconds to finish querying out all records
* Update Lucene Index Job with better timings and timeouts
* Adjust package index fields.
* Don't force a full rebuild every package addition/update
* Store all unlisted as those should also be returned in some instances
* Move FindPackagesById to use the index when storing all results
* Fix the search filters a bit with the id, tags, author precursors

ferventcoder:
There were a number of issues identified in the move to store all results in the index (#754). Most of these are addressed by this request:

* Set timeout for connection itself to allow for more than 110  seconds to finish querying out all records
* Update Lucene Index Job with better timings and timeouts
* Adjust package index fields.
* Don't force a full rebuild every package addition/update
* Store all unlisted as those should also be returned in some instances
* Move FindPackagesById to use the index when storing all results
* Fix the search filters a bit with the id, tags, author precursors
* Ensure index is updated properly when a package is updated and storing all results

Fixes #761
Fixes #762

### DIFF
--- a/chocolatey/Website/Entities/EntitiesContext.cs
+++ b/chocolatey/Website/Entities/EntitiesContext.cs
@@ -39,15 +39,43 @@ namespace NuGetGallery
 
     public class EntitiesContext : DbContext, IWorkItemsContext, IEntitiesContext
     {
-        public EntitiesContext(): base(SetConnectionString("NuGetGallery"))
+        private const int DEFAULT_TIMEOUT_SECONDS = 110;
+
+        public EntitiesContext() : this("NuGetGallery", DEFAULT_TIMEOUT_SECONDS)
         {
-            InitializeCustomOptions();
         }
 
-        private static string SetConnectionString(string connectionName)
+        public EntitiesContext(int timeoutSeconds) : this("NuGetGallery", timeoutSeconds)
+        {
+        } 
+        
+        public EntitiesContext(string nameOrConnectionString) : this(nameOrConnectionString,DEFAULT_TIMEOUT_SECONDS)
+        {
+        } 
+        
+        public EntitiesContext(string nameOrConnectionString, int timeoutSeconds)
+            : base(SetConnectionString(nameOrConnectionString))
+        {
+            InitializeCustomOptions(timeoutSeconds);
+        }
+
+        private static bool TreatAsConnectionString(string nameOrConnectionString)
+        {
+            return nameOrConnectionString.IndexOf('=') >= 0;
+        }
+
+        private static string SetConnectionString(string nameOrConnectionString)
         {
             var connectionString = new StringBuilder();
-            connectionString.Append(System.Configuration.ConfigurationManager.ConnectionStrings[connectionName].ConnectionString);
+
+            if (!TreatAsConnectionString(nameOrConnectionString))
+            {
+                connectionString.Append(System.Configuration.ConfigurationManager.ConnectionStrings[nameOrConnectionString].ConnectionString);
+            }
+            else
+            {
+                connectionString.Append(nameOrConnectionString);
+            }
 
             if (!connectionString.to_string().EndsWith(";"))
             {
@@ -65,7 +93,7 @@ namespace NuGetGallery
         /// <summary>
         ///   Initializes the custom options.
         /// </summary>
-        protected void InitializeCustomOptions()
+        protected void InitializeCustomOptions(int timeoutSeconds)
         {
             // defaults for quick reference
             //Configuration.LazyLoadingEnabled = true;
@@ -79,7 +107,7 @@ namespace NuGetGallery
             if (adapter != null)
             {
                 var objectContext = adapter.ObjectContext;
-                objectContext.CommandTimeout = 110; // value in seconds
+                objectContext.CommandTimeout = timeoutSeconds; // value in seconds
             }
        }
 
--- a/chocolatey/Website/Infrastructure/Lucene/LuceneIndexingService.cs
+++ b/chocolatey/Website/Infrastructure/Lucene/LuceneIndexingService.cs
@@ -35,12 +35,14 @@ namespace NuGetGallery
 {
     public class LuceneIndexingService : IIndexingService
     {
+        private const int LUCENE_CACHE_REBUILD_HOURS = 24;
+        private const int LUCENE_UPDATE_JOB_FREQUENCY_MINUTES = 10;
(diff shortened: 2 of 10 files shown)


chocolatey/Website/Entities/EntitiesContext.cs
```diff
@@ -39,15 +39,43 @@ public interface IEntitiesContext
 
     public class EntitiesContext : DbContext, IWorkItemsContext, IEntitiesContext
     {
-        public EntitiesContext(): base(SetConnectionString("NuGetGallery"))
+        private const int DEFAULT_TIMEOUT_SECONDS = 110;
+
+        public EntitiesContext() : this("NuGetGallery", DEFAULT_TIMEOUT_SECONDS)
```
ferventcoder: I thought about pulling this out to a const as it is used in two places now in this file.

chocolatey/Website/Entities/EntitiesContext.cs
```diff
@@ -39,15 +39,43 @@ public interface IEntitiesContext
...
         {
-            InitializeCustomOptions();
+            InitializeCustomOptions(timeoutSeconds);
         }
 
-        private static string SetConnectionString(string connectionName)
+        private static bool TreatAsConnectionString(string nameOrConnectionString)
+        {
+            return nameOrConnectionString.IndexOf('=') >= 0;
+        }
```
ferventcoder: This is code lifted from EF that this code base didn't have access to call. Being able to do this allows a fix in case a full connection string is passed in instead of using a connection string from a config file.

chocolatey/Website/Infrastructure/Lucene/LuceneIndexingService.cs
```diff
@@ -35,25 +35,31 @@ namespace NuGetGallery
 {
     public class LuceneIndexingService : IIndexingService
     {
+        private const int LUCENE_CACHE_REBUILD_HOURS = 24;
+        private const int LUCENE_UPDATE_JOB_FREQUENCY_MINUTES = 10;
+        private const int LUCENE_UPDATE_JOB_TIMEOUT_MINUTES = 8;
```
ferventcoder: These remove the magic numbers from before. The previous values were 3, 6, and 5, respectively.

chocolatey/Website/Infrastructure/Lucene/LuceneIndexingService.cs
```diff
@@ -134,37 +153,45 @@ public void UpdatePackage(Package package)
...
-                    p => (_indexContainsAllVersions || p.IsLatest || p.IsLatestStable) &&
-                         p.PackageRegistration.Packages.Any(p2 => p2.LastUpdated > lastIndexTime));
-            }
-            else if (!_indexContainsAllVersions)
-            {
-                set = set.Where(p => p.IsLatest || p.IsLatestStable); // which implies that p.IsListed by the way!
-            }
+                var packageRepo = new EntityRepository<Package>(context);
 
-            set = set.Where(p => p.Listed);
```
ferventcoder: Most of this area is items being moved under a contextThunk function invoke, but this bit here was removed as we actually do want everything that is not rejected, or 

` set = set.Where(p => p.StatusForDatabase != _rejectedStatus  || p.StatusForDatabase == null);`

as you'll see at line 173ish.

chocolatey/Website/Infrastructure/Lucene/LuceneSearchService.cs
```diff
@@ -75,9 +76,9 @@ private SearchResults SearchCore(SearchFilter searchFilter)
 
             if (searchFilter.IncludeAllVersions)
             {
-                filter = new QueryWrapperFilter(new TermQuery(new Term("Listed", Boolean.TrueString)));
+                filter = new QueryWrapperFilter(new TermQuery(new Term("InIndex", Boolean.TrueString)));
```
ferventcoder: Added InIndex always set to true as we needed a filter term for the below. We just needed something that would always evaluate to true.

chocolatey/Website/Infrastructure/Lucene/LuceneSearchService.cs
```diff
@@ -288,18 +285,23 @@ private static Query ParseQuery(SearchFilter searchFilter)
...
             }
-            
+
             // Create an OR of all the queries that we have
             var combinedQuery = conjuctionQuery.Combine(new Query[] { exactIdQuery, relatedIdQuery, exactTitleQuery, startsIdQuery, startsTitleQuery, wildCardIdQuery, wildCardTitleQuery, conjuctionQuery, wildCardQuery });
 
-            if (onlySearchById)
+            if (onlySearchByExactId)
+            {
+                combinedQuery = conjuctionQuery.Combine(new Query[] { exactIdQuery });
```
ferventcoder: This allows for FindPackagesById to work

chocolatey/Website/Services/PackageService.cs
```diff
@@ -1204,7 +1204,7 @@ private PackageOwnerRequest FindExistingPackageOwnerRequest(PackageRegistration
 
         private void NotifyIndexingService()
         {
-            indexingSvc.UpdateIndex(forceRefresh:true);
+            indexingSvc.UpdateIndex(forceRefresh:false);
```
ferventcoder: This is likely a case of younger Rob thinking this was a good idea.

chocolatey/Website/Infrastructure/Lucene/LuceneIndexingService.cs
```diff
@@ -101,6 +108,14 @@ public void UpdateIndex(bool forceRefresh)
 
         public void UpdatePackage(Package package)
         {
+            if (_indexContainsAllVersions)
+            {
+                //just update everything since the last write time
+                UpdateIndex(forceRefresh: false);
+                return;
```
ferventcoder: This fixes our duplicates by updating all package versions in the index, so only one is marked as latest and shows up properly.

PR #784, opened 2019-12-30
(GH-783) Fix broken link to moderator queue

### LINKED ISSUE #783: Going to `https://chocolatey.org/packages?q=&moderatorQueue=true` breaks
### What You Are Seeing?
If a user tries to go to https://chocolatey.org/packages?q=&moderatorQueue=true the website breaks. This link is used in a few areas of the website.

### What is Expected?
I would expect that link to work, and to be taken to the Moderator Queue

Go to https://chocolatey.org/packages?q=&moderatorQueue=true and notice the "Oh no, we broke something!" page.

designxsteph:
Currently, if a user is to go to https://chocolatey.org/packages?q=&moderatorQueue=true they will be hit with a 404 page.

This is due to the moderationStatus filter being null. Now, if the filter is null then it will be defaulted to "AllModerationStatuses" which will show all the packages in moderation as expected. With this in place, the link above now works as intended.

Fixes #783

### DIFF
--- a/chocolatey/Website/Controllers/PackagesController.cs
+++ b/chocolatey/Website/Controllers/PackagesController.cs
@@ -399,7 +399,13 @@ namespace NuGetGallery
                 // If we are searching for something, sort by relevance.
                 sortOrder = q.IsEmpty() ? Constants.PopularitySortOrder : Constants.RelevanceSortOrder;
             }
-            
+
+            if (String.IsNullOrEmpty(moderationStatus))
+            {
+                // If no moderation status is specified, default to All Statuses
+                moderationStatus = Constants.AllModerationStatuses;
+            }
+
             int totalHits = 0;
             int updatedPackagesCount = 0;
             int respondedPackagesCount = 0;


chocolatey/Website/Controllers/PackagesController.cs
```diff
@@ -399,7 +399,13 @@ public virtual ActionResult ListPackages(string q, string sortOrder = null, int
                 // If we are searching for something, sort by relevance.
                 sortOrder = q.IsEmpty() ? Constants.PopularitySortOrder : Constants.RelevanceSortOrder;
             }
-            
+
+            if (String.IsNullOrEmpty(moderationStatus))
```
vexx32: I usually see most code linters recommend that the `string` reserved name is used rather than the `String` type name for static method calls.

Not really sure why, but I happened to notice it. 🤷‍♂ 

```suggestion
            if (string.IsNullOrEmpty(moderationStatus))
```

AdmiringWorm: I believe it is more of a preference kind of thing than anything else.
People seem to like to use the built-in aliases instead of the concrete class/struct types, I assume because the aliases are easier to type since they are in all lowercase and sometimes shorter as well (I could be wrong here of course).

vexx32: Yeah it strikes me as a preference thing by and large. [This SO answer](https://stackoverflow.com/questions/452368/why-do-lowercase-and-uppercase-versions-of-string-exist-and-which-should-i-use) has a few more details.

Probably worth pointing out that using the alias is "safer" in some sense -- `String` as a type name could potentially register as ambiguous if you happen to (unwisely) define a type by the same name. I don't think there's a way to override the builtin `string` alias though, so that should be safe (happy to be proven wrong there though, seems there's always something weird you can do on occasion 😂).

AdmiringWorm: > I don't think there's a way to override the builtin string alias though, so that should be safe (happy to be proven wrong there though, seems there's always something weird you can do on occasion 😂).

I assume you would get a compile time error if you tried to redefine one of the aliases. Probably possible with IL manipulation (No idea how though 😄)

PR #786, opened 2019-12-31
(GH-785) Database optimizations

### LINKED ISSUE #785: Database Query Optimizations
After being reviewed closer by folks smarter than myself, we've come up with the following additional optimizations.

- Further reduce nvarchar(max) on fields that don't need to have it (follows up #747).
- ~~Add an Index on Packages (Listed, IsPrerelease) INCLUDE (PackageRegistrationKey, Description, Summary)~~ existing index should be sufficient?
- ~~Rewrite moderation queue query (packages list) to only return the fields that it needs.~~ (we'll hold on this for now as it would be rewriting the entire packages page display logic to use a different query)
- Don't use LOWER when searching case insensitive database

ferventcoder:
Fixes #785


chocolatey/Website/Migrations/201912302335265_AdjustVarCharMaxColumnsPartTwo.cs
```diff
@@ -0,0 +1,219 @@
...
+)
+
+PRINT 'Dropping ' + @ConstraintName
+EXEC('ALTER TABLE [dbo].[Users] DROP CONSTRAINT ' + @ConstraintName)");
+   
+
+            AlterColumn("Users", "EmailAddress", c => c.String(maxLength: 150));
+            AlterColumn("Users", "UnconfirmedEmailAddress", c => c.String(maxLength: 150));
+            AlterColumn("Users", "HashedPassword", c => c.String(maxLength: 256));
+            AlterColumn("Users", "PasswordHashAlgorithm", c => c.String(maxLength: 20, nullable: false)); // has a defaultValue: "SHA1", so we need to remove that and readd it separately.
```
gep13:
```suggestion
            AlterColumn("Users", "PasswordHashAlgorithm", c => c.String(maxLength: 20, nullable: false)); // has a defaultValue: "SHA1", so we need to remove that and read it separately.
```

mkevenaar: @gep13 I think it might be read as re-add

gep13: Ah, now that I read it properly, I think you are right, I read it as a typo initially. I think your suggestion makes sense though.

ferventcoder: re-add is what I meant here.

Issue #790, opened 2020-01-02
4 spaces being converted to code block on package description


designxsteph:
### What You Are Seeing?
On packages that have 4 spaces inside the package description area, the 4 spaces is being converted to a code block when it should not be. 

See: https://chocolatey.org/packages/powershell-preview#description and compare it to the nuspec at: https://github.com/DarwinJS/ChocoPackages/blob/d6967a91b3787020f4938aa0f8a8face8d47e134/pwshpackages/powershell-preview/powershell-preview.nuspec#L20-L48

### What is Expected?
In this case it is not intended for the 4 spaces to be converted to a code block.

### How Did You Get This To Happen? (Steps to Reproduce)
1. Go to https://chocolatey.org/packages/powershell-preview#description and notice that the description is all in a code block and highlighted.

![code-block](https://user-images.githubusercontent.com/42750725/71671654-355e9c00-2d39-11ea-844e-6118786d8ee4.png)

gep13:
@designxsteph this has been deployed...

![image](https://user-images.githubusercontent.com/1271146/71982387-40736980-321c-11ea-929e-c71269b33cc8.png)

PR #791, opened 2020-01-02
(GH-790) Prevent 4 spaces from converting to code blocks

designxsteph:
In the package description area, currently 4 spaces is being interpreted as a code block when that is the unintended outcome. The changes here will disable this feature. This is implemented on both the Verify Package page and the Display Package page.

Reference on a similar issue: https://github.com/lunet-io/markdig/issues/348 
Package in pictures: https://chocolatey.org/packages/powershell-preview#description

Before:
code-block

After:
code-block-after

### DIFF
--- a/chocolatey/Website/Views/Packages/DisplayPackage.cshtml
+++ b/chocolatey/Website/Views/Packages/DisplayPackage.cshtml
@@ -1,5 +1,6 @@
 ﻿@using Markdig
 @using NuGetGallery
+@using Markdig.Parsers
 @model DisplayPackageViewModel
 @{
     Layout = "~/Views/Shared/NewsletterLayout.cshtml";
@@ -15,7 +16,7 @@
     Bundles.Reference("Scripts/easymde");
     Bundles.Reference("Scripts/closeable");
 
-    var MarkdownPipeline = new MarkdownPipelineBuilder()
+    var markdownPipelineBuilder = new MarkdownPipelineBuilder()
                  .UseSoftlineBreakAsHardlineBreak()
                  .UseAutoLinks()
                  .UseGridTables()
@@ -24,8 +25,11 @@
                  .UseEmphasisExtras()
                  .UseNoFollowLinks()
                  .UseCustomContainers()
-                 .UseBootstrap()
-                 .Build();
+                 .UseBootstrap();
+
+    markdownPipelineBuilder.BlockParsers.TryRemove<IndentedCodeBlockParser>();
+
+    var MarkdownPipeline = markdownPipelineBuilder.Build();
 
     var statuses = Model.Status.GetEnumerationItems().ToList();
     var packageImage = Url.ImageUrl(Model.Id, Model.Version, Model.IconUrl);
--- a/chocolatey/Website/Views/Packages/VerifyPackage.cshtml
+++ b/chocolatey/Website/Views/Packages/VerifyPackage.cshtml
@@ -1,5 +1,6 @@
 ﻿@using Markdig
-@using NuGetGallery;
+@using NuGetGallery
+@using Markdig.Parsers
 @model VerifyPackageViewModel
 @{
     ViewBag.Tab = "Upload Your Package";
@@ -10,7 +11,7 @@
     Bundles.Reference("Scripts/packages");
     Bundles.Reference("Scripts/prism");
 
-    var MarkdownPipeline = new MarkdownPipelineBuilder()
+    var markdownPipelineBuilder = new MarkdownPipelineBuilder()
                  .UseSoftlineBreakAsHardlineBreak()
                  .UseAutoLinks()
                  .UseGridTables()
@@ -19,8 +20,11 @@
                  .UseEmphasisExtras()
                  .UseNoFollowLinks()
                  .UseCustomContainers()
-                 .UseBootstrap()
-                 .Build();
+                 .UseBootstrap();
+
+    markdownPipelineBuilder.BlockParsers.TryRemove<IndentedCodeBlockParser>();
+
+    var MarkdownPipeline = markdownPipelineBuilder.Build();
 }
 
 <section id="account" class="container py-3 py-md-5">
--- a/chocolatey/Website/Website.csproj
+++ b/chocolatey/Website/Website.csproj
@@ -147,8 +147,8 @@
     <Reference Include="Lucene.Net.Contrib.SpellChecker">
       <HintPath>..\..\packages\Lucene.Net.Contrib.3.0.3\lib\net40\Lucene.Net.Contrib.SpellChecker.dll</HintPath>
     </Reference>
-    <Reference Include="Markdig, Version=0.15.6.0, Culture=neutral, processorArchitecture=MSIL">
-      <HintPath>..\..\packages\Markdig.0.15.6\lib\net40\Markdig.dll</HintPath>
+    <Reference Include="Markdig, Version=0.18.0.0, Culture=neutral, processorArchitecture=MSIL">
+      <HintPath>..\..\packages\Markdig.0.18.0\lib\net40\Markdig.dll</HintPath>
     </Reference>
     <Reference Include="MarkdownSharp, Version=2.0.0.0, Culture=neutral, processorArchitecture=MSIL">
       <HintPath>..\..\packages\MarkdownSharp.2.0.5\lib\net40\MarkdownSharp.dll</HintPath>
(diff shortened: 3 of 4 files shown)


designxsteph:
While this approach does seem to work, I'm not sure if this is how it should have been handled or not, and open to suggestions. Markdig, the markdown converter we are using, uses CommonMark guidelines. Based off CommonMark guidelines 4 spaces is a valid way of making a code block.

See: https://commonmark.org/help/
Markdig: https://github.com/lunet-io/markdig

So, in turn it is not an added extension of Markdig causing this issue, it is the built in functionality of Markdig. I have found a resource that may be helpful in modifying this built in feature of Markdig here https://github.com/lunet-io/markdig/blob/7964bd0160d4c18e4155127a4c863d61ebd8944a/src/Markdig/MarkdownExtensions.cs#L306 but am unsure if this is more correct than what I have already implemented, or how to apply it to this situation (not converting 4 spaces to code blocks).

chocolatey/Website/Controllers/PackagesController.cs
```diff
@@ -1141,5 +1143,18 @@ private string FindAndCorrectInvalidMarkdownHeaders(string markdownText)
                 (match => match.Groups[1].Value + " " + match.Groups[2].Value),
                 RegexOptions.Multiline);
         }
+
+        private string FindAndRemoveSpacesCodeBlocks(string codeBlock)
+        {
+            var spacesCodeBlockPattern = @"(^ {4})(\S.*)";
```
vexx32: Hmm. I guess the main question here is do we want to _completely_ disable the `4-spaces -> codeblock` functionality for Markdig?

It would certainly simplify things if we did. Otherwise, we have to figure out how we want to handle edge cases:

### 1. "intentional"(?) code blocks

Consider a description formatted like this:

```md
    <start description>

        <some other paragraph>

    <continue description>
```

The current regex will catch the start and end, but leave the middle untouched to be later formatted as a code block. Is that what we want?

### 2. Excessive but consistent indentation

Consider:

```md
        line 1
        line 2
        line 3
```

This is sourced from the package's nuspec XML, right? Folx could well opt to format it differently (XML is usually whitespace insensitive, so it's not usually a problem to adjust the formatting of the nuspec) and leave us with this becoming a codeblock, even though it might just be something like this in the XML:

```xml
<?xml>
<package>
    <description>
        line 1
        line 2
        line 3
    </description>
</package>
```

With this regex, this whole thing would still end up formatted as a code block, wouldn't it?

---

If we adjust the regex, we can effectively disable Markdig's 4-space-code-block formatter by simply removing any multiple of 4 spaces from the start of any line:

```
(^ {4})+(\S.*)
```

However, this carries a problematic byproduct -- within, say, a fenced codeblock, we also run the risk of stripping deliberate indentation.

Not sure what a full solution would look like, but it might end up being a bit more complex, so this might be more complicated to handle than trying to iron out a few edge cases. 🙂

chocolatey/Website/Controllers/PackagesController.cs
```diff
@@ -171,6 +171,7 @@ public virtual ActionResult DisplayPackage(string id, string version)
             package.Files = packageSvc.GetPackageFiles(package).ToList();
             package.Description = FindAndCorrectInvalidMarkdownHeaders(package.Description);
             package.ReleaseNotes = FindAndCorrectInvalidMarkdownHeaders(package.ReleaseNotes);
+            package.Description = FindAndRemoveSpacesCodeBlocks(package.Description);
```
ferventcoder: Please combine with line 172, nesting these items:

~~~c#
package.Description = FindAndRemoveSpacesCodeBlocks(FindAndCorrectInvalidMarkdownHeaders(package.Description));
~~~

chocolatey/Website/Controllers/PackagesController.cs
ferventcoder: > I guess the main question here is do we want to completely disable the 4-spaces -> codeblock functionality for Markdig?

Yes, and my question on seeing this is if there is a setting in Markdig that will just shut that off? If not, it might be worth asking them about it (Markdig folks).

chocolatey/Website/Controllers/PackagesController.cs
```diff
@@ -189,6 +190,7 @@ public virtual ActionResult DisplayPackage(string id, string version, FormCollec
             package.Files = packageSvc.GetPackageFiles(package, useCache: false).ToList();
             package.Description = FindAndCorrectInvalidMarkdownHeaders(package.Description);
             package.ReleaseNotes = FindAndCorrectInvalidMarkdownHeaders(package.ReleaseNotes);
+            package.Description = FindAndRemoveSpacesCodeBlocks(package.Description);
```
ferventcoder: Same here as above, nest.

chocolatey/Website/Controllers/PackagesController.cs
gep13: Where is the right place to document that we are not going to support 4 spaces as code blocks? I don’t think we have anywhere currently that would make sense. Perhaps a new documentation page is required?

ferventcoder: Actually, someone already asked this question - @designxsteph please see https://github.com/lunet-io/markdig/issues/348#issuecomment-503591458

ferventcoder: @gep13 we never supported it before for descriptions and release notes.  (speaking of @designxsteph, this would need to apply to releasenotes as well).

ferventcoder: > Where is the right place to document that we are not going to support 4 spaces as code blocks? I don’t think we have anywhere currently that would make sense. Perhaps a new documentation page is required?

@gep13 I think a documentation page would be great on what parts of Markdown we don't support. Would you like to write up an issue please?

AdmiringWorm: > we never supported it before for descriptions and release notes

While perhaps not being supported, code blocks using 4 space indendation have been available for at least the past 3-4 years (yes even with the old markdownsharp library).
I remember getting hit by this when I started out creating chocolatey packages.

gep13: Added an issue here: https://github.com/chocolatey/chocolatey.org/issues/792

designxsteph:
I have updated this PR to using the method described here https://github.com/lunet-io/markdig/issues/348

chocolatey/Website/Views/Packages/DisplayPackage.cshtml
```diff
@@ -15,7 +16,7 @@
     Bundles.Reference("Scripts/easymde");
     Bundles.Reference("Scripts/closeable");
 
-    var MarkdownPipeline = new MarkdownPipelineBuilder()
+    var markdownPipeline = new MarkdownPipelineBuilder()
```
vexx32: If we're using this in two places already, would it be worth pulling this method of building a Markdown pipeline into a new method? That way, we have one central method that we can modify as needed, so we can be sure we're always using the same Markdown parsing settings?

designxsteph: I agree that would be better, but before when we were using MarkdownSharp still it was being done in the cshtml themselves (like it is now). I had tried to move it the controller, but when doing that the `clean_html` property was causing blocks of code to be re-rendered and appended to multiple times, but only reload after a form was submitted. Really strange! I'm thinking that's why it was initially done in the cshtml. So, that is another issue that would need to be worked out.

ferventcoder: We might e able to pull it into a helpers or something so all can call the same thing, and then call it from here so there is just one line that is repeated (aka `var MarkdownPipeline = SomeHelper.SetAndBuildMarkdownPipeline();`). That's outside this PR though.

chocolatey/Website/Controllers/PackagesController.cs
ferventcoder: > I remember getting hit by this when I started out creating chocolatey packages.

Yep, and we fixed that pretty quickly to not do that.

ferventcoder: We had our own custom MarkdownSharp library implementation, and that was one of the things we removed.

ferventcoder:
Has this been deployed out to test? Would like to get this up at some point!

designxsteph:
@ferventcoder This is on test now!

designxsteph:
> So I do have one change I need to request - variables should not differ only by case. Please rename the builder to markdownPipelineBuilder so the name doesn't match the MarkdownPipeline variable.

Gotchya, this has been updated.

gep13:

gep13:
Fixes #790

Issue #793, opened 2020-01-02
Visibility Enhancements


ferventcoder:
There are some enhancements we should add to provide more visibility when we are in tracking or looking at issues.

PR #794, opened 2020-01-02
(GH-793) Visibility Enhancements

ferventcoder:
* Connect background jobs under different user
* Report on invalid search filter
* Restrict non-cached calls with search filter modifiers
* Trace execution time for Lucene and certain repositories

### DIFF
--- a/chocolatey/Website/DataServices/FeedServiceBase.cs
+++ b/chocolatey/Website/DataServices/FeedServiceBase.cs
@@ -166,6 +166,8 @@ namespace NuGetGallery
                 return GetResultsFromSearchService(searchFilter);
             }
 
+            Trace.WriteLine("Search filter was invalid ('{0}') Raw Url: '{1}' .".format_with(searchFilter.FilterInvalidReason, HttpContext.Request.RawUrl));
+
             if (!includePrerelease)
             {
                 packages = packages.Where(p => !p.IsPrerelease);
--- a/chocolatey/Website/DataServices/FeedServiceBase.cs
+++ b/chocolatey/Website/DataServices/FeedServiceBase.cs
@@ -29,6 +29,7 @@ using System.ServiceModel;
 using System.Text.RegularExpressions;
 using System.Web;
 using System.Web.Mvc;
+using Elmah;
 using NuGetGallery.MvcOverrides;
 using QueryInterceptor;
 
@@ -166,7 +167,10 @@ namespace NuGetGallery
                 return GetResultsFromSearchService(searchFilter);
             }
 
-            Trace.WriteLine("Search filter was invalid ('{0}') Raw Url: '{1}' .".format_with(searchFilter.FilterInvalidReason, HttpContext.Request.RawUrl));
+            var invalidSearchFilterMessage = "Search filter was invalid ('{0}') Raw Url: '{1}' .".format_with(searchFilter.FilterInvalidReason, HttpContext.Request.RawUrl);
+            Trace.WriteLine(invalidSearchFilterMessage);
+            // raise this as an exception for even better visibility
+            ErrorSignal.FromCurrentContext().Raise(new SystemException(invalidSearchFilterMessage));
 
             if (!includePrerelease)
             {
--- a/chocolatey/Website/DataServices/FeedServiceBase.cs
+++ b/chocolatey/Website/DataServices/FeedServiceBase.cs
@@ -173,36 +173,6 @@ namespace NuGetGallery
                 packages = packages.Where(p => !p.IsPrerelease);
             }
 
-            //if (searchFilter.Skip != 0)
-            //{
-            //    packages = packages.Skip(searchFilter.Skip);
-            //}
-
-            //packages = packages.Take(MaxPageSize);
-            //if (searchFilter.Take != 0)
-            //{
-            //    packages = packages.Take(searchFilter.Take);
-            //}
-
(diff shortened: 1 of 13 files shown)


chocolatey/Website/Entities/EntitiesContext.cs
```diff
@@ -59,6 +60,41 @@ public EntitiesContext(string nameOrConnectionString, int timeoutSeconds)
...
+            }
+
+            if (!hasReplacedPassword || !hasReplacedUser)
+            {
+                Trace.WriteLine("Error adjusting connection string - username and password not found to replace.");
+            }
+
+            return returnedConnectionString.to_string();
+        }
+
```
ferventcoder: Agree that there is a better way to handle this - which is to have two connection strings. However our current provider doesn't provide the ability to edit two connection strings, so best to take the current and just change it over to the new user/password.

chocolatey/Website/Entities/EntityRepository.cs
```diff
@@ -57,8 +90,16 @@ public IQueryable<T> GetAll()
 
         public int InsertOnCommit(T entity)
         {
-            entities.Set<T>()
-                    .Add(entity);
+            if (TraceLogEvents)
+            {
+                MethodExtensionWrappers.WrapExecutionTracingTime(
+                    () => entities.Set<T>().Add(entity),
+                    () => "Added new '{0}' ('{1}') to database".format_with(typeof(T).Name, entity.Key.to_string()));
```
ferventcoder: Note this is deferred so that the entity should have a Key set by the time this message comes across.

ferventcoder: should...

chocolatey/Website/Infrastructure/Lucene/LuceneIndexingService.cs
```diff
@@ -166,29 +169,41 @@ private List<PackageIndexEntity> GetPackages(DateTime? lastIndexTime)
...
-                var packagesForIndexing = list.Select(
-                    p => new PackageIndexEntity
+                var list = MethodExtensionWrappers.WrapExecutionTracingTime(() =>
+                {
+                   return set
+                            .Include(p => p.PackageRegistration)
+                            .Include(p => p.PackageRegistration.Owners)
+                            .Include(p => p.SupportedFrameworks)
+                            .ToList();
+                },"Lucene Indexer: Select from database completed");
```
ferventcoder: This is expected to take the longest

chocolatey/Website/MethodExtensionsWrappers.cs
```diff
@@ -0,0 +1,116 @@
...
+                    Trace.WriteLine(message() + " (" + (totalTime / TimeSpan.TicksPerMillisecond) + " milliseconds)");
+                }
+                else
+                {
+                    Trace.WriteLine(message() + " (" + (totalTime / TimeSpan.TicksPerSecond) + " seconds)");
+                }
+            }
+        }
+    }
+}
```
ferventcoder: I explored trying to add these in as extensions to the lambda functions, however it wouldn't work with anonymous methods. It required me to create a variable for the function and then I could call extension methods on that. That wasn't going to add to readability, so I backed this down to wrappers.

ferventcoder:
~~~c#
var list = () =>
                {
                   return set
                            .Include(p => p.PackageRegistration)
                            .Include(p => p.PackageRegistration.Owners)
                            .Include(p => p.SupportedFrameworks)
                            .ToList();
                }.WrapExecutionTracingTime("Lucene Indexer: Select from database completed");
~~~

Is what I wanted to do for readability instead of:

~~~c#
var list = MethodExtensionWrappers.WrapExecutionTracingTime(() =>
                {
                   return set
                            .Include(p => p.PackageRegistration)
                            .Include(p => p.PackageRegistration.Owners)
                            .Include(p => p.SupportedFrameworks)
                            .ToList();
                },"Lucene Indexer: Select from database completed");
~~~

but the only way it would work is if I first declared the function (pseudocode):

~~~c#
Func<IQueryable<Package>> getPackages = () =>
                {
                   return set
                            .Include(p => p.PackageRegistration)
                            .Include(p => p.PackageRegistration.Owners)
                            .Include(p => p.SupportedFrameworks)
                            .ToList();
                };
var list = getPackages..WrapExecutionTracingTime("Lucene Indexer: Select from database completed");
~~~

I guess it's not terrible on readability now that I look at it, but it kind of defeated the purpose of what I wanted to do here.

chocolatey/Website/DataServices/FeedServiceBase.cs
```diff
@@ -166,41 +167,16 @@ protected virtual IQueryable<Package> SearchCore(IQueryable<Package> packages, s
                 return GetResultsFromSearchService(searchFilter);
             }
 
+            var invalidSearchFilterMessage = "Search filter was invalid ('{0}') Raw Url: '{1}' .".format_with(searchFilter.FilterInvalidReason, HttpContext.Request.RawUrl);
+            Trace.WriteLine(invalidSearchFilterMessage);
+            // raise this as an exception for even better visibility
+            ErrorSignal.FromCurrentContext().Raise(new SystemException(invalidSearchFilterMessage));
```
ferventcoder: For even more visibility, treat and log this as an error. This will allow review from the app itself.

PR #808, opened 2020-01-12
Enable TLS 1.2 & 1.3 Icon Downloading

### LINKED ISSUE #652: chocolatey.org Is unable to cache icons from websites requiring TLS 1.2
### What You Are Seeing?
When a maintainer is using an icon from a website that requires TLS 1.2 to be used, chocolatey.org is unable to cache that icon on the server.
Which results in a broken icon on the package.

### What is Expected?
Expects chocolatey.org to be able to cache and show the icon even if the website the icon comes from requires TLS 1.2.

Upload a new package version with a iconUrl from a TLS 1.2 only webpage (only one I know of right now is github raw links, but could be others).

### Possible Solutions
1. Update the caching code to set the service point to allow TLS 1.2 (this will require .NET 4.5 or higher to be installed on the machine compiling the website, and on the server itself).
2. If the image isn't cached locally, use the specified icon url as it is specified in the nuspec file.
3. Implement a custom downloader that allows for TLS 1.2

joel0:
Some CDNs and iconUrl servers have disabled TLS versions below 1.2, causing problems for the WebClient in .NET 4.0 (see reports about the Statically CDN failing for icons).

Targeting a newer version of .NET should should remove the need for this, but for now this should fix #652 .

I've tested this change on Windows 10 with the Statically CDN to confirm that it resolved the problem.

### DIFF
--- a/chocolatey/Website/App_Start/AppActivator.cs
+++ b/chocolatey/Website/App_Start/AppActivator.cs
@@ -22,6 +22,7 @@ using System.Configuration;
 using System.Data.Entity;
 using System.Data.Entity.Migrations;
 using System.Linq;
+using System.Net;
 using System.Web;
 using System.Web.Mvc;
 using System.Web.Routing;
@@ -50,6 +51,19 @@ namespace NuGetGallery
         public static void PreStart()
         {
             MiniProfilerPreStart();
+
+            // Enable TLS 1.3, TLS 1.2, TLS 1.1, and TLS 1.0 for all WebClients. Specifically, for
+            // Services\ImageFileService.cs#DownloadImage() to allow downloading package icons served from CDNs that
+            // have disabled TLS <1.2 from .NET 4.0. This also disables SSLv3, if it was enabled by default.
+            SecurityProtocolType enabledProtocols = 0;
+            foreach (int protocol in new int[] { 12288, 3072, 768, 192 })
+            {
+                if (Enum.IsDefined(typeof(SecurityProtocolType), protocol))
+                {
+                    enabledProtocols |= (SecurityProtocolType)protocol;
+                }
+            }
+            ServicePointManager.SecurityProtocol = enabledProtocols;
         }
 
         public static void PostStart()


ferventcoder:
@joel0 I don't see what issue this is associated with - can we get these commits associated to an issue? Thanks!

ferventcoder:
I see in your commit you added #652. However we have a specific way of structuring commit messages here. Can I have you review https://github.com/chocolatey/choco/blob/master/CONTRIBUTING.md#prepare-commits and then update your PR?

gep13:
@joel0 did you mean to close this PR?  That shouldn't be necessary to make the requested changes.

joel0:
@gep13  I did not mean to.  I did a force push, which closed it.  I'm trying to figure out how to fix this and re-open this PR.

joel0:
I have rewritten the commit message to align with the contributing guidelines.  Sorry for not reading those earlier.

chocolatey/Website/App_Start/AppActivator.cs
```diff
@@ -50,6 +51,17 @@ public static class AppActivator
         public static void PreStart()
         {
             MiniProfilerPreStart();
+
+            // Enable TLS 1.3, TLS 1.2, TLS 1.1, and TLS 1.0 for all WebClients. Specifically, for
+            // Services\ImageFileService.cs#DownloadImage() to allow downloading package icons served from CDNs that
+            // have disabled TLS <1.2 from .NET 4.0.
+            try
+            {
+                ServicePointManager.SecurityProtocol = (SecurityProtocolType)(12288 | 3072 | 768 | 192);
```
ferventcoder: Can you look at https://github.com/chocolatey/chocolatey.org/pull/654/files and emulate this aspect a bit to handle not having TLS 1.3 installed?

ferventcoder: Otherwise I see this possibly erroring if someone does not have .NET 4.8+ installed.

joel0:
Thanks for the feedback.  I have pushed a fix, which also avoids throwing and catching exceptions altogether.

PR #821, opened 2020-01-24
Merge pull request #821 from designxsteph/GH-731

designxsteph:
This removes unknown packages from the total moderation count and from
being viewed from the "all-statuses" filter while in the moderation
queue. This has been done to reduce confusion on the total number of
packages in moderation. These packages are still visible but only by
following the url below:

https://chocolatey.org/packages?q=&moderatorQueue=true&moderationStatus=unknown-status

Adds to #731


gep13:

Issue #823, opened 2020-01-28
Null cookie key values causing ArgumentNullException


designxsteph:
### What You Are Seeing?
An `ArgumentNullException` is being thrown on the `CookieTempDataProvider.cs` file at lines 35 and 71. The website appears to be running fine until the logs are inspected. Seems that there is nothing to guard against these null values. It can happen by visiting https://chocolatey.org/packages?q=tag%3Acli

### What is Expected?
I would expect that `CookieTempDataProvider.cs` would guard against/handle null values.

PR #824, opened 2020-01-28
Merge pull request #824 from designxsteph/null-cookie

designxsteph:
In `CookieTempDataProvider.cs` there was nothing to protect against null key values. Although the website appeared to be running fine, this was causing `ArgumentNullException` errors to be thrown. Parts of the file `CookieTempDataProvider.cs` has been updated to the latests code which includes protection against this and handles null values appropriately. The `secure` flag was also added to the cookie as good practice (and is also in the updated file).

Reference:
https://github.com/NuGet/NuGetGallery/blob/master/src/NuGetGallery/Infrastructure/CookieTempDataProvider.cs


gep13:
Fixes #823

gep13:

PR #825, opened 2020-01-30
 (GH-818) Prevent CSRF attack on Logout

gep13:
The changes here update the logoff action to a `HttpPost` instead of an
`HttpGet` which allows the usage of an antiforgery token to prevent any
CSRF attacks on logout.


gep13:
Relates to #818

PR #826, opened 2020-01-30
(GH-829) Add dispose-and-retry to index commits

vexx32:
Lucene docs recommend handling OutOfMemoryException by disposing the IndexWriter. See here: https://lucenenetdocs.azurewebsites.net/api/Lucene.Net/Lucene.Net.Index.IndexWriter.html

This should prevent index corruption and allow us to simply re-create the index writer and retry the commit. If the second commit attempt fails, then the error will be raised. This should hopefully give us some breathing room in high load situations.

Related: #829 

/cc @ferventcoder @gep13

### DIFF
--- a/chocolatey/Website/Infrastructure/Lucene/LuceneIndexingService.cs
+++ b/chocolatey/Website/Infrastructure/Lucene/LuceneIndexingService.cs
@@ -24,6 +24,7 @@ using System.Diagnostics;
 using System.Globalization;
 using System.IO;
 using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
 using Elmah;
 using Lucene.Net.Index;
@@ -84,12 +85,28 @@ namespace NuGetGallery
 
                 if ((lastWriteTime == null) || IndexRequiresRefresh() || forceRefresh)
                 {
-                    EnsureIndexWriter(creatingIndex: true);
-                    Trace.WriteLine("Lucene Index: Deleting index");
-                    _indexWriter.DeleteAll();
-                    Trace.WriteLine("Lucene Index: Index delete completed");
-                    _indexWriter.Commit();
-                    Trace.WriteLine("Lucene Index: Index delete committed");
+                    void handleOutOfMemory()
+                    {
+                        _indexWriter.Dispose();
+                        _indexWriter = null;
+
+                        EnsureIndexWriter(creatingIndex: true);
+                    }
+
+                    void nukeIndex()
+                    {
+                        EnsureIndexWriter(creatingIndex: true);
+                        Debug.Assert(_indexWriter != null);
+
+                        Trace.WriteLine("Lucene Index: Deleting index");
+                        _indexWriter.DeleteAll();
+                        Trace.WriteLine("Lucene Index: Index delete completed");
+
+                        _indexWriter.Commit();
+                        Trace.WriteLine("Lucene Index: Index delete committed");
+                    }
+
+                    DoAndRetryOnOutOfMemory(nukeIndex, handleOutOfMemory);
 
                     // Reset the lastWriteTime to null. This will allow us to get a fresh copy of all the packages
                     lastWriteTime = null;
@@ -131,26 +148,34 @@ namespace NuGetGallery
                     {
                         var packageRepo = new EntityRepository<Package>(context);
                         package = packageRepo.GetAll()
-                                    .Where(p => (p.IsLatest || p.IsLatestStable) && p.PackageRegistrationKey == packageRegistrationKey)
-                                    .Include(p => p.PackageRegistration)
-                                    .Include(p => p.PackageRegistration.Owners)
-                                    .Include(p => p.SupportedFrameworks)
-                                    .FirstOrDefault();
+                            .Where(p => (p.IsLatest || p.IsLatestStable) && p.PackageRegistrationKey == packageRegistrationKey)
+                            .Include(p => p.PackageRegistration)
+                            .Include(p => p.PackageRegistration.Owners)
+                            .Include(p => p.SupportedFrameworks)
+                            .FirstOrDefault();
                     }
                 }
 
-                // Just update the provided package
+                void handleOutOfMemory() => DisposeAndEnsureIndexWriter(creatingIndex: false);
+
                 EnsureIndexWriter(creatingIndex: false);
+
+                // Just update the provided package
                 if (package != null)
                 {
-                    var indexEntity = new PackageIndexEntity(package);
-                    _indexWriter.UpdateDocument(updateTerm, indexEntity.ToDocument());
+                    var document = new PackageIndexEntity(package).ToDocument();
+
+                    DoAndRetryOnOutOfMemory(
+                        () => _indexWriter.UpdateDocument(updateTerm, document),
+                        handleOutOfMemory,
+                        waitMilliseconds: 50);
                 }
                 else
                 {
-                    _indexWriter.DeleteDocuments(updateTerm);
+                    DoAndRetryOnOutOfMemory(() => _indexWriter.DeleteDocuments(new Term[] { updateTerm }), handleOutOfMemory, waitMilliseconds: 50);
                 }
-                _indexWriter.Commit();
+
+                DoAndRetryOnOutOfMemory(() => _indexWriter.Commit(), handleOutOfMemory);
             }
         }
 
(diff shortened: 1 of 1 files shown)


gep13:
@vexx32 is there an issue associated with this PR?  If not, I think we need to create one, so that we can hook up the commits to the issue.

vexx32:
#613 is the closest I can find. I'll open up a new one for now so we can track it as we go.

gep13:
>@vexx32 said...
#613 is the closest I can find. I'll open up a new one for now so we can track it as we go.

I believe that that is a different issue, as it only affects when pushing embedded packages.  This issue affects all update operations against the site.

vexx32:
Created #829 to track this. 🙂

gep13:
@vexx32  Within the LuceneIndexingService we are using the following methods on the IndexWriter:

- DeleteAll
- UpdateDocument
- DeleteDocuments
- Commit
- AddDocument

According to the documentation that you pointed out, each of these methods "could" throw an Out Of Memory Exception.  Do you think we should add safe methods for each of these?

vexx32:
@gep13 yeah, I think that's probably a good call. I wasn't sure it'd be necessary, but it does make sense to at least have some code in place to handle it if/when it happens. I can give this a quick look tomorrow and add some methods for those cases as well. 🙂

chocolatey/Website/Infrastructure/Lucene/LuceneIndexingService.cs
```diff
@@ -231,7 +232,34 @@ private void AddPackagesCore(IList<PackageIndexEntity> packages, bool creatingIn
...
+        // Per https://lucenenetdocs.azurewebsites.net/api/Lucene.Net/Lucene.Net.Index.IndexWriter.html we should expect and handle OutOfMemoryException
+        private void SafeIndexCommit(bool creatingIndex)
+        {
+            try
+            {
+                _indexWriter.Commit();
+            }
+            catch (OutOfMemoryException)
+            {
+                _indexWriter.Dispose();
```
ferventcoder: Does this mean that the index that has been set is destroyed? Or does the work on the local index that has been done stick around?

chocolatey/Website/Infrastructure/Lucene/LuceneIndexingService.cs
```diff
@@ -231,7 +232,34 @@ private void AddPackagesCore(IList<PackageIndexEntity> packages, bool creatingIn
...
+                _indexWriter = null;
+
+                EnsureIndexWriter(creatingIndex: creatingIndex);
```
ferventcoder: Following up on the previous question, this would delete all of the work we are trying to save with the first commit, no?

chocolatey/Website/Infrastructure/Lucene/LuceneIndexingService.cs
vexx32: The index that is currently present on disk is called back to the last known good state, before the commit operation was begun, to prevent corruption of the stored index. That's about as much as the docs say.

chocolatey/Website/Infrastructure/Lucene/LuceneIndexingService.cs
vexx32: Hmm. Yes, possibly. But given we're handling an out of memory exception, I'm not sure there's a more robust avenue available to us at that point. I suppose I could wrap this up differently, into a `TryCommit()` sort of method and look to retry the operation from further up the stack.

vexx32: I'll take a closer look at this and @gep13's question tomorrow morning, and see if I can find a more thorough way to go.

vexx32: Re-checked the stacktrace we had from Elmah, and yeah we're hitting this when we try to `UpdatePackage()` which subsequently updates the index. So I think doing a `TryCommit()` in there is probably the right thing to do, with possibly a couple retries before throwing some kind of exception.

vexx32: (Among other places that we need to call Commit() of couse.)

vexx32: God, a `Try...()` that catches an OOM exception seems **horrifically** cursed, but c'est la vie.

vexx32:
@ferventcoder @gep13 Reworked the solution to include anything that calls `Commit()` and allow those methods to retry the operation they're attempting. Going to add a new commit or two to address the other potentially problematic methods that @gep13 mentioned.

gep13:
@vexx32 can you update your commit messages to include the GitHub issue that they are associated with?

gep13:
@vexx32 sorry to be a pain, but the format that is normally used is the following:

https://github.com/chocolatey/choco/commits/master

gep13:
You can see the same in action here:

https://github.com/chocolatey/chocolatey.org/commits/master

vexx32:
No worries, all sorted. 🙂

chocolatey/Website/Infrastructure/Lucene/LuceneIndexingService.cs
ferventcoder: So in other words, the changes won't commit. Which means things will be wrong until the next index rebuild.

chocolatey/Website/Infrastructure/Lucene/LuceneIndexingService.cs
```diff
@@ -224,14 +257,155 @@ private void AddPackagesCore(IList<PackageIndexEntity> packages, bool creatingIn
...
+                _indexWriter.Dispose();
+                _indexWriter = null;
+
+                storedException = e;
+                return false;
+            }
+        }
+
+        // Per https://lucenenetdocs.azurewebsites.net/api/Lucene.Net/Lucene.Net.Index.IndexWriter.html we should expect and handle OutOfMemoryException
+        private bool TryDeleteDocuments(Term[] terms, out Exception storedException)
```
ferventcoder: This is private, but still needs a guard on null reference exceptions for terms.

chocolatey/Website/Infrastructure/Lucene/LuceneIndexingService.cs
```diff
@@ -224,14 +257,155 @@ private void AddPackagesCore(IList<PackageIndexEntity> packages, bool creatingIn
...
+                return false;
+            }
+            catch (OutOfMemoryException e)
+            {
+                storedException = e;
+                return false;
+            }
+        }
+        // Per https://lucenenetdocs.azurewebsites.net/api/Lucene.Net/Lucene.Net.Index.IndexWriter.html we should expect and handle OutOfMemoryException
+        private bool TryUpdateDocument(Term term, Lucene.Net.Documents.Document document, out Exception storedException)
```
ferventcoder: All items here (minus the known null storedException) should have a guard clause to assert not null

chocolatey/Website/Infrastructure/Lucene/LuceneIndexingService.cs
```diff
@@ -139,18 +152,38 @@ public void UpdatePackage(Package package)
...
+                    {
+                        // Commit succeeded, exit loop and return.
+                        return;
+                    }
+
+                    retryCount++;
+                } while (retryCount < 3);
+
+                // If we get here, all retries have failed, rethrow last failure exception.
+                throw updateFailException;
```
ferventcoder: Instead of repeating retry logic in multiple places, let's instead implement https://github.com/chocolatey/choco/blob/564a65235ea9b019fe0abbf3f9929b98da49ce5c/src/chocolatey/infrastructure/tolerance/FaultTolerance.cs#L52-L66 in this code base.

chocolatey/Website/Infrastructure/Lucene/LuceneIndexingService.cs
```diff
@@ -84,18 +84,31 @@ public void UpdateIndex(bool forceRefresh)
...
                 if ((lastWriteTime == null) || IndexRequiresRefresh() || forceRefresh)
                 {
-                    EnsureIndexWriter(creatingIndex: true);
-                    Trace.WriteLine("Lucene Index: Deleting index");
-                    _indexWriter.DeleteAll();
-                    Trace.WriteLine("Lucene Index: Index delete completed");
-                    _indexWriter.Commit();
-                    Trace.WriteLine("Lucene Index: Index delete committed");
+                    int retryCount = 0;
+                    Exception deleteFailedException;
```
ferventcoder: I'm not sure how I feel about passing around an exception as an out variable for methods. It feels wrong - exception handling has a known pattern where we can do that. I imagine this was to have logic in one place for handling and retry, but that can be done with a retry policy codeblock.

chocolatey/Website/Infrastructure/Lucene/LuceneIndexingService.cs
ferventcoder: Sorry, this is the full set - https://github.com/chocolatey/choco/blob/564a65235ea9b019fe0abbf3f9929b98da49ce5c/src/chocolatey/infrastructure/tolerance/FaultTolerance.cs#L44-L119

chocolatey/Website/Infrastructure/Lucene/LuceneIndexingService.cs
ferventcoder: It's good etiquette to let the reviewer resolve the conversation. :D

vexx32: Oh, this probably automatically resolved / was marked outdated (thanks Github lol) because I updated it.

chocolatey/Website/Infrastructure/Lucene/LuceneIndexingService.cs
vexx32: I think that makes sense. I considered doing something like this, but wasn't sure how to do it properly in C#. In PowerShell it's a lot easier 😂

chocolatey/Website/Infrastructure/Lucene/LuceneIndexingService.cs
vexx32: Not entirely sure what you're recommending here to be quite honest. Can we set up a meeting so we can go through this a bit? 🙂

chocolatey/Website/Infrastructure/Lucene/LuceneIndexingService.cs
```diff
@@ -219,19 +244,109 @@ public void AddPackages(IList<PackageIndexEntity> packages, bool creatingIndex)
...
+                        handleException.Invoke();
+                    }
+
+                    Thread.Sleep(waitMilliseconds + (attempts * increaseWaitOnRetryMilliseconds));
+                    Trace.WriteLine("Retrying action.");
+                }
+            }
+
+            Trace.WriteLine("Retryable action succeeded.");
+            return result;
```
ferventcoder: I'm okay with these not being broken out to its own class at this point. If it looks to be used in other places, I would see this moved out.

vexx32: Thankfully the Lucene madness is pretty self-contained. 😉

Issue #827, opened 2020-01-30
JS errors on Install and Verify Package pages


designxsteph:
### What You Are Seeing?
On both the Verify Package (step 2 of uploading a package) page and on the Install page, there is a JS error of `'querySelectorAll' of undefined`. On the Install page, this is causing the install steps area not to work, where users type in their URL of choice and it is inserted into the correct area of the designated install script below. 

### What is Expected?
I would expect these 2 pages not to have JS errors and for all to be working as it should. 

### How Did You Get This To Happen? (Steps to Reproduce)
1. Visit https://chocolatey.org/install
2. Open up developer tools and notice the JS error
3. Type a URL in into Step 4, and notice it does not change in the designated install scripts area.

PR #828, opened 2020-01-30
(GH-827) Fix JS on Install and Verify Package page

designxsteph:
Currently when going to the Install or Verify Package (step 2 when uploading a package) page, there is a JS error of `'querySelectorAll' of undefined`, which is being caused by a snippet of JS that is meant to be run specifically on the Package Display page.

This error is causing the install steps area not to work on the Install page, and the file that contains this code is actually not needed at all on the Verify Package page.

The JS snippet that controls the install steps has been moved to its own JS file, which allows the `package-details.js` to be removed entirely from the 2 pages that do not need it. This cuts down on the amount of unnecessary JS on these 2 pages. The `install-steps.js` file has been added to the Package Display page and the Install page as those are the 2 pages that require it at this time.

In result, the file that is causing the initial error has been removed and contained to the specific page that is needed on which is the Package Details page.

### DIFF
--- a/chocolatey/Website/Scripts/install-steps/install-steps.js
+++ b/chocolatey/Website/Scripts/install-steps/install-steps.js
@@ -1,0 +1,20 @@
+$(function () {
+    $("code:contains('STEP 3 URL')").html(function (_, html) {
+        $(this).parent().next().remove();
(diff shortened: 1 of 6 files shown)


gep13:

gep13:
Fixes #827

PR #836, opened 2020-02-04
(GH-835, GH-838) Get username on authentication via ajax and prevent noscript flicker

### LINKED ISSUE #835: Profiles link in sub navigation on first page of packages does not work
### What You Are Seeing?
On the first page of the [package list](https://chocolatey.org/packages) the `profile` link does not work when logged in. This link does work though on every other page except this one.

### What is Expected?
I would expect the link to work no matter what page I am viewing it on.

1. Go to https://chocolatey.org/packages
2. Log in to your account
3. Click on the `profile` link in the sub navigation from the first page of https://chocolatey.org/packages

### LINKED ISSUE #838: Noscript message sometimes flashes in top navigation
### What You Are Seeing?
On some pages, where the authentication sub navigation is present, there is a flicker of the `noscript` message on the screen before the JavaScript has a chance to remove it. This is a minor issue, but one that can become a nuisance.

### What is Expected?
I would expect to not see the flashing of the noscript text on load.

1. Log into your account
2. Go to your account page and watch the top navigation. You may see a flicker of red text.

designxsteph:
The first page of the package area `/packages` is heavily cached. For this reason, we need to get the username via ajax call in order for the `profiles` link in the sub navigation to work. Now that we have the username, JavaScript can inject it on load, so that the link will work even if the page is heavily cached.

Method of retrieving the username in `PagesController.cs` is similar to how it is implemented in `UsersController.cs`.

This PR also prevents the flashing of the `noscript` text on load in the authentication sub navigation.

Fixes #835
Fixes #838

### DIFF
--- a/chocolatey/Website/Controllers/PagesController.cs
+++ b/chocolatey/Website/Controllers/PagesController.cs
@@ -20,6 +20,7 @@ using System;
 using System.Linq;
 using System.Net.Mail;
 using System.Net.Mime;
+using System.Security.Principal;
 using System.Text;
 using System.Text.RegularExpressions;
 using System.Web.Mvc;
@@ -32,11 +33,14 @@ namespace NuGetGallery
     {
         private readonly IAggregateStatsService statsSvc;
         private readonly IMessageService messageService;
+        private readonly IPrincipal currentUser;
 
-        public PagesController(IAggregateStatsService statsSvc, IMessageService messageService)
+        public PagesController(IAggregateStatsService statsSvc, IMessageService messageService, IPrincipal currentUser)
         {
             this.statsSvc = statsSvc;
             this.messageService = messageService;
+            this.currentUser = currentUser;
+
         }
 
         [HttpGet, OutputCache(CacheProfile = "Cache_2Hours")]
@@ -681,7 +685,17 @@ SET PATH=%PATH%;%ALLUSERSPROFILE%\chocolatey\bin";
         [ActionName("Home")]
         public virtual ActionResult IsAuthenticatedAjax()
         {
-            return Json(new { isAuthenticated = @Request.IsAuthenticated }, JsonRequestBehavior.AllowGet);
+            if (Request.IsAuthenticated)
+            {
+                return Json(
+                    new
+                    {
+                        isAuthenticated = true,
+                        userName = currentUser.Identity.Name
+                    }, JsonRequestBehavior.AllowGet);
+            }
+
+            return Json(new { isAuthenticated = false }, JsonRequestBehavior.AllowGet);
         }
     }
 }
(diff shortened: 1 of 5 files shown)


chocolatey/Website/Controllers/PagesController.cs
```diff
@@ -681,7 +685,11 @@ public virtual JsonResult Stats()
         [ActionName("Home")]
         public virtual ActionResult IsAuthenticatedAjax()
         {
-            return Json(new { isAuthenticated = @Request.IsAuthenticated }, JsonRequestBehavior.AllowGet);
+            return Json(
+                new {
+                    isAuthenticated = Request.IsAuthenticated,
+                    userName = currentUser.Identity.Name
```
gep13: @designxsteph what happens here when the user isn't authenticated?  In that situation, I would expect either the `currentUser` or `Identity` to be null, and as a result, throw an exception.  We might want to alter this slightly to only return an object if `Request.IsAuthenticated`

designxsteph: @gep13 This has been updated. If there's a better approach other than how I did it let me know!

chocolatey/Website/Controllers/PagesController.cs
```diff
@@ -681,7 +685,17 @@ public virtual JsonResult Stats()
         [ActionName("Home")]
         public virtual ActionResult IsAuthenticatedAjax()
         {
-            return Json(new { isAuthenticated = @Request.IsAuthenticated }, JsonRequestBehavior.AllowGet);
+            if (Request.IsAuthenticated)
+            {
+                return Json(
+                    new
+                    {
+                        isAuthenticated = Request.IsAuthenticated,
```
gep13: @designxsteph here we don't need to access the property again, we could just say `true`

chocolatey/Website/Controllers/PagesController.cs
```diff
@@ -681,7 +685,17 @@ public virtual JsonResult Stats()
...
+            {
+                return Json(
+                    new
+                    {
+                        isAuthenticated = Request.IsAuthenticated,
+                        userName = currentUser.Identity.Name
+                    }, JsonRequestBehavior.AllowGet);
+            }
+
+            return Json(new { isAuthenticated = Request.IsAuthenticated }, JsonRequestBehavior.AllowGet);
```
gep13: @designxsteph here we don't need to access the property again, we could just say `false`

chocolatey/Website/Controllers/PagesController.cs
designxsteph: Good point. This has been updated.

chocolatey/Website/Controllers/PagesController.cs
designxsteph: Good point. This has been updated.

designxsteph:
@gep13 Since it's both related to the authentication sub navivation, I added one more commit to this PR that will prevent the random flickering of the `noscript` text.

Issue #853, opened 2020-04-01
NullReferenceException and AlreadyClosed errors from Index writer accesses


vexx32:
See (#829) for prior reference.

Following the related PR's merge (#826) and code pushed to the website, things have looked _more_ stable but still prone to failure after a few days.

Current errors seen in the logs related to the index writer include:

- NullReferenceException
- AlreadyClosed

This indicates we may have threading issues, and some index writes may still be lost while the index writer is being regenerated.

Possible fix may be to block get operations for the index writer while it is being regenerated. I'm poking at the code currently, may submit a PR in a bit for review.

/cc @gep13  and @ferventcoder

ferventcoder:
Merged into master at https://github.com/chocolatey/chocolatey.org/commit/1f70af7bdeb70482965351d28eac571a1adba604. Pushed out just now.

PR #854, opened 2020-04-01
(#853) Lock index writer when recovering from errors.

vexx32:
## Summary

Currently we're hitting some issues where accessing the index writer can cause errors indicating it has been closed, or just straight up NullReferenceExceptions.

It's likely these occur as a result of accessing the `_indexWriter` from multiple threads, and because we're not properly removing it from the WriterCache dictionary. This PR should ensure that requests for the index writer block during the recovery sequence, so that it has time to be recreated properly, and the reference to the broken writer in the cache is removed correctly.

## Changes

+ Created new Index property and updated all _indexWriter references.

+ Lock in the getter for Index so requests block during recovery.

+ Removed some redundant code in `EnsureIndexWriter` that was evading the only lock we had already in place.

+ Ensure we remove the broken index writer from the dictionary cache, so we avoid offering up the disposed writer once again.


ferventcoder:
@vexx32 This needs to be rebased.

vexx32:
@ferventcoder done!

PR #855, opened 2020-04-02
Remove Top Alert Only on "X" Click and Maintenance

designxsteph:
Currently, the top alert banner will be removed once a user clicks on
it by using a session storage item. Now, the top alert banner will stay
in place unless the user specifically clicks on the "X" in the right
corner to remove it.

Minor updates to the JS snippet controlling this feature was made. The
changes also eliminates the flash of red sometimes seen on page load if
a user has removed the alert.

Also includes:
* Typo fix
* Adds suggested video from Part 1 video


designxsteph:
@ferventcoder @gep13 Whenever this is pushed, the cache on the main Scripts file will most likely need purged for this change to take effect.

designxsteph:
@ferventcoder @gep13 Added two more small commits, this is ready to go.

PR #857, opened 2020-04-08
(layout) Adjust CSS on Profile Page

designxsteph:
Due to recent changes, the CSS that controls custom list styling has
been been moved to it's own scss file. The small snippet from
`_base.css` was missed during this transition and is now causing a
styling error on the profile page.

The snippet has been moved to the appropriate file which makes the
profile page appear how it should.


gep13:
@designxsteph thanks for fixing this!

PR #863, opened 2020-04-17
(layout) Add Styling for Blockquotes

designxsteph:
Adds styling for blockquote elements to resemble the usage of "callouts"
throughout the website. This will mostly be used when converting files
from markdown, otherwise the standard "callout" syntax should be used.


designxsteph:
@ferventcoder or @gep13 the custom.js file cache will likely need purged to allow this change to take effect.

ferventcoder:
and pushed

PR #890, opened 2020-06-04
(layout) Adjust Home Page Announcements

designxsteph:
Updates the announcements on the home page to unite the sections on deployments and updates the webinar to the new Deployments Webinar.


designxsteph:
This is on test now

Issue #798, opened 2020-01-06
500 HTTP error on uploading a package


taoxin2988:
### What You Are Seeing?
Oh no, we broke something!
You might be here because:
There is an error on our server and the server is "given her all she's got, captain."
The system is down, the system is down, the system is down...

### What is Expected?
I would  like to know how long the function of edit package will be restored

### How Did You Get This To Happen? (Steps to Reproduce)
 when I upload a package and click "Verify Details & Submit", I got  en error upon in the reply for several days.

pauby:
You don't mention which package(s) this is for?

taoxin2988:
> You don't mention which package(s) this is for?

my package is platon_mtool_other, the current version is 0.7.4, and I want to update to 0.7.5

benshep:
I have this issue too, for getiplayer updating from 3.23 to 3.24.

pauby:
@taoxin2988 Thanks for clarifying that - I wanted to check the [package exists](https://chocolatey.org/packages/platon_mtool_other). To get the error message you are clicking the 'Edit Package' button on the package page and then you immediately get the 'Oh no, we broke something!'. Are those the exact steps you're taking?

pauby:
@benshep Can you clarify exactly what you're doing to get that error?

taoxin2988:
> @benshep Can you clarify exactly what you're doing to get that error?

Actually, you can operates like this: first of all,  go into Account; second, click "Upload Package", and you will be asked to upload your  ****.nupkg , next you need to click "Verify Details & Submit", then I got the error above.

taoxin2988:
> @taoxin2988 Thanks for clarifying that - I wanted to check the [package exists](https://chocolatey.org/packages/platon_mtool_other). To get the error message you are clicking the 'Edit Package' button on the package page and then you immediately get the 'Oh no, we broke something!'. Are those the exact steps you're taking?

alternatively, you can click "Manage Packages" in your Account, and select a package you wanna update, then you will find a button "Edit Package" in that page,click this button and you will link into the upload pages like what I descripted above.

benshep:
> 
> 
> @taoxin2988 Thanks for clarifying that - I wanted to check the [package exists](https://chocolatey.org/packages/platon_mtool_other). To get the error message you are clicking the 'Edit Package' button on the package page and then you immediately get the 'Oh no, we broke something!'. Are those the exact steps you're taking?

Yes, that's exactly what I'm doing.

[...]

PR #891, opened 2020-06-05
(GH-798) Remove ability to upload nupkg 

gep13:
This is known to cause a number of problems, and the recommendation has
been for a long time for folks to use the choco push command.  Removing
the upload through the site encourages people to use the recommended
way.

The page structure has been left in place in case anyone has shortcuts
to the path, and information has been left on that page for the correct
course of action.

Relates to #798


chocolatey/Website/Views/Packages/UploadPackage.cshtml
```diff
@@ -3,77 +3,23 @@
 @{
     ViewBag.Tab = "Upload Your Package";
     Bundles.Reference("Content/dist/chocolatey.slim.css");
-    Bundles.Reference("Content/account.css");
+    Bundles.Reference("Content/prism/prism.css");
```
ferventcoder, line 6: Little lost on this change (along with the prism below)

gep13: I had a chat with @designxsteph while doing this PR. The prism addition is for the styling of the text box around the code snippet, and while adding that, she mentioned that the account.css is actually not required, so took the opportunity to remove it.

ferventcoder:
It may be worth exploring a web upload component at some point in the future, but it's best to consider this as a way forward that works.

PR #892, opened 2020-06-05
(maint) Update Timers on Event Pages

designxsteph:
The countdown timers on the event pages were off due to not being
converted to the correct timezone. Now, the `EventDate` is set to UTC
across the board to ensure consistency. This field is only used to set
the countdown clocks on the individual event, and to move items on the
main event page from "upcoming" to the "on-demand" sections.


designxsteph:
The main /Scripts file cache will need to be purged with the deployment of this PR.

PR #893, opened 2020-06-05
(GH-865) Limit number of package versions allowed in a submitted status

### LINKED ISSUE #865: Don't allow new package versions to be pushed if there are X number not in a final state
Recently, there was over 100 new package versions pushed in one day, the majority of those over 3 different packages which overwhelmed the package scanner.

I feel we should consider limiting pushes of new package versions if there are x number not in an _final_ state (ie. Approved or Rejected).

ferventcoder:
There should be no more than a certain number of versions of a package in moderation at one time. If there are, it means that under normal circumstances communications are not working appropriately.

Closes #865


ferventcoder:
Fixed up the grammar and missed number item in the error message.

Issue #897, opened 2020-06-19
Install.ps1 throwing an error if pscx module is installed


joshcorr:
### What You Are Seeing?
Error when installing Chocolatey.
```
Querying latest package from https://proget.internal.repo/nuget/chocolatey/Packages()?$filter=(Id%20eq%20%27chocolatey%27)%20and%20IsLatestVersion
Downloading  to c:\temp\chocolatey.0.10.15.nupkg
No file exists at c:\temp\chocolatey.0.10.15.nupkg
At line:7 char:9
+         Invoke-Command -ScriptBlock { & C:\temp\Chocolatey.ps1} -Sess ...
+         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : OperationStopped: (No file exists ...y.0.10.15.nupkg:String) [], RuntimeException
    + FullyQualifiedErrorId : No file exists at c:\temp\chocolatey.0.10.15.nupkg
 
A parameter cannot be found that matches parameter name 'DestinationPath'.
The term 'C:\Users\<userName>\AppData\Local\Temp\chocolatey\chocInstall\tools\chocolateyInstall.ps1' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
```
### What is Expected?
Chocolatey is installed via the Script.

### How Did You Get This To Happen? (Steps to Reproduce)

Was using the [Offline Install](https://chocolatey.org/courses/installation/installing?method=completely-offline-install) script with the native `Expand-Archive` to setup some nodes  and we had the Pscx module installed on some nodes, which caused the error above.
It looks like the Public Script may also have this bug.

Believe this could be fixed by using the full path to `Expand-Archive` `Microsoft.PowerShell.Archive\Expand-Archive`

PR #898, opened 2020-06-20
(GH-897) Fully Qualified Expand-Archive

joshcorr:
This is a proposed change for #897

Please reach out if you have any questions about this issue.


gep13:
@vexx32 @steviecoaster @pauby any thoughts/concerns on this change?

gep13:
**NOTE:** If this PR is merged, it will need the signature block to be updated.

vexx32:
Makes sense to do, I think, if only because the pscx module was pretty popular for a while and may still persist in some environments.

Seems like a good change to me. :+1:

joshcorr:
This may need to be added to other scripts or documentation locations like [this one](https://chocolatey.org/docs/installation#completely-offline-install). I originally found the problem when installing from a local nuget repo.

joshcorr:
@ferventcoder @gep13 @vexx32 Did this PR need to get pulled in? Or Closed since #912 was doing a similar thing?

vexx32:
@joshcorr closing this for now as I've incorporated it into the more extensive changes in #912. Apologies for the long wait here, we'll get it sorted soon though 😁 

Thanks for your work here!

PR #908, opened 2020-07-14
(GH-909) Use wildcard when including files into csproj file

### LINKED ISSUE #909: Use wildcard within csproj file to include all files
Currently, when a new documentation file is brought over from the choco wiki, you have to remember to go into the csproj file and include this new file into it, otherwise the new piece of documentation doesn't get included.

This happened recently, and had to be updated:

https://github.com/chocolatey/chocolatey.org/commit/5ff175979aeb0a2143e4f6a6c859f2faf3cf5ee0

This has also happened in the past when creating new blog posts for the site.  Blog posts are typically created outside of Visual Studio, so the addition to the csproj file isn't made automatically.

When testing locally, since the files are on disk, everything looks like it is working properly, but when published, the pages aren't included, since the files haven't been copied to the correct location.

gep13:
This will prevent folks, myself included, from forgetting to update the csproj file when adding new files into the website.

Fixes #909


gep13:
With these changes in place, files are still all included in the Blog folder:

image

as well as in the Files folder:

image

gep13:
This same technique can likely be used in a couple other places as well, but these two are the main ones where I know that there have been issues with an update to the csproj file being forgotten.

PR #910, opened 2020-07-17
Add package scanner into moderation process

### LINKED ISSUE #877: Trusted packages that have X+ number of virus detections should be held for moderation
I'm seeing trusted packages that have a high Virus Total detection count (as shown on the package page) in an approved state.

Trusted packages that have a higher number than X detections for 1 or more files should be held back for human moderation. The X needs to be agreed. My suggestion is 5.

gep13:
There are still some things that need to be done before this PR can be merged in, but it is ready for a review.

I have successfully walked a package through the happy path, i.e. submit package, pass validator, pass verifier, pass scanner, and then have the package approved by a moderator.  I have also tested this when a package is trusted.  Things that still need to be done.

- [ ] When a error occurs on the package page, i.e. when attempting to approve a package, the validation summary text is in small black text, and hard to see.  This needs to be changed.
- [ ] We should use a modal popup when something goes wrong, to indicate there is a problem.  Similar to what is shown when it updates successsfully
- [ ] We need to decide on exactly what causes a package to be pushed back to the maintainer, and when it goes to human moderation.  At the minute, a package is never pushed back to a maintainer based on scanner results
- [ ] Add input textbox on package page for validation exemption message
- [ ] Add input textbox on package page for scanner exemption message
- [ ] Add link in status summary section to the virus scan results.  Similar to the `Details` hyperlink shown for verification testing URL
- [ ] Show any files that were not scanned by the package-scanner (**NOTE:** This may have to be broken out into a separate scope of work
- [ ] Need to update disclaimer about virus scanning only happening on en-US locale

Fixes #877


gep13:
Here is the happy path...

Package submitted to chocolatey.org...

image

Package passes validator...

image

Package passes verifier...

image

Package passes scanner and is approved (this is a trusted package)...

image

gep13:
Validation Summary styling has been re-applied.  Now looks like this:

image

When there is a problem

gep13:
An error modal will now be shown when something went wrong with the requested operation:

image

chocolatey/Website/Views/Packages/DisplayPackage.cshtml
```diff
@@ -559,6 +562,50 @@
...
+                                    <p>At least one file within this package had greater then 0 detections for a file</p>
+                                }
+                                else if (Model.PackageScanFlagResultType == PackageScanFlagResultType.Warning)
+                                {
+                                    <p>At least one file within this package had between 5 and 10 detections for a file</p>
+                                }
+                                else if (Model.PackageScanFlagResultType == PackageScanFlagResultType.Error)
+                                {
+                                    <p>At least one file within this package had greater then 10 detections for a file</p>
+                                }
```
vexx32, line 585: Can this be written as a `switch` statement? I get the impression it would be clearer than this stack of if/elseif

chocolatey/Website/Views/Packages/DisplayPackage.cshtml
```diff
@@ -559,6 +562,50 @@
...
+                            <p class="mb-0"><span class="fas fa-question text-secondary mr-3"></span> Scan Testing Unknown</p>
+                        }
+                        @if (Model.PackageScanResultStatus == PackageScanStatusType.Exempted)
+                        {
+                            <span class="fas fa-ban text-secondary mr-3"></span>
+                            <div>
+                                <p class="mb-0">Scan Testing Exempt:</p>
+                                @Html.Raw(Markdown.ToHtml(Model.ExemptedFromScannerReason.clean_html() ?? string.Empty, MarkdownPipeline))
+                            </div>
+                        }
```
vexx32, line 607: These statuses are mutually exclusive, right? Should these be `else if`s instead?

Might also be a good candidate for a switch statement.

chocolatey/Website/Controllers/ApiController.cs
```diff
@@ -553,6 +553,21 @@ public virtual ActionResult SubmitPackageScanResults(string apiKey, string id, s
...
 
+            PackageScanFlagResultType packageScanFlagResult;
+            try
+            {
+                Enum.TryParse(scanFlagResult.to_string(), true, out packageScanFlagResult);
+            }
+            catch (Exception)
+            {
+                packageScanFlagResult = PackageScanFlagResultType.Unknown;
+            }
```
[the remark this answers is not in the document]
AdmiringWorm: Only if the out parameter isn't of an enum type.
Would probably be better to replace the try/catch statements with for instance.
```suggestion
            if (!Enum.TryParse(scanFlagResult.to_string(), true, out packageScanFlagResult))
            {
                packageScanFlagResult = PackageScanFlagResultType.Unknown;
            }
```

ferventcoder:
> Validation Summary styling has been re-applied. Now looks like this:
> 
> image
> 
> When there is a problem

This is a different issue IMHO

chocolatey/Website/Controllers/ApiController.cs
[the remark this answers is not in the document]
ferventcoder: > Would probably be better to replace the try/catch statements with for instance.

If it comes back with true/false (like one would expect it to as it is a "Try" method), then this adjustment would be appropriate. However if you give it null/empty/not a string representation of the enum, and it can throw an error, then your suggested change is incorrect.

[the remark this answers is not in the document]
AdmiringWorm: > However if you give it null/empty/not a string representation of the enum, and it can throw an error, then your suggested change is incorrect.

I agree, if that had been the case trying to parse without a try/catch block would be incorrect. Thankfully, that isn't the case though, and since the `to_string` extension method also does a null check of the object, there shouldn't be a problem there either.

enum-parsing

gep13:
>@ferventcoder said...
This is a different issue IMHO - can it get its own issue number?

I have created associated issues #916 and #917

chocolatey/Website/Content/packages.css
```diff
@@ -136,4 +136,28 @@ textarea {
...
+}
+
+.validation-summary-errors ul li {
+    padding-left: 1rem;
+    border-left: 6px solid #f4516c;
+}
+
+.validation-summary-errors ul li:not(:first-child) {
+    margin-top: .5rem;
+}
```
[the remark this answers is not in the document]
gep13: I will need to defer to @designxsteph on this one.  I unashamedly stole the css that I have put in here from this file: https://github.com/chocolatey/chocolatey.org/blob/master/chocolatey/Website/Content/account.css#L32-L50

I suspect that during a refactoring it was removed, so I have placed it into the package.css file, which is already included on the page.

chocolatey/Website/Controllers/ApiController.cs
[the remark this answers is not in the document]
gep13: This addition was based on the code that is already in place for parsing exisitng enums, so I adapted the same logic for this new enum: https://github.com/chocolatey/chocolatey.org/blob/master/chocolatey/Website/Controllers/ApiController.cs#L607-L615

chocolatey/Website/Content/packages.css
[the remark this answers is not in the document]
vexx32: Fair enough! I tend to be a little wary of using `:not()` clauses and such, especially in combination with other pseudo-classes. Might just be a pet peeve, but I have a vague recollection that they're slower for browsers to process/render (though I'm sure not by a great deal, so... eh).

chocolatey/Website/Views/Packages/DisplayPackage.cshtml
gep13: Perhaps it can, but I am a big fan of consistency in the code base, and currently, this is how the other two sections of the code base are currently implemented: https://github.com/chocolatey/chocolatey.org/blob/master/chocolatey/Website/Views/Packages/DisplayPackage.cshtml#L511-L560

chocolatey/Website/Controllers/ApiController.cs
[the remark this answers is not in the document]
vexx32: Perhaps best left as a refactoring issue to be dealt with separately, then? Either way we should probably cut the unneeded try/catches, no sense maintaining unreachable code paths. 🙂

chocolatey/Website/Views/Packages/DisplayPackage.cshtml
gep13: :+1: This has been changed.

chocolatey/Website/Views/Packages/DisplayPackage.cshtml
vexx32: I agree that consistency is valuable, but IMO it's better to not waste the compute cycles needlessly re-checking things repeatedly that have already satisfied the only case they're going to satisfy.

Up to you, we can always do a pass over all the cases like this in a follow up PR or something.

chocolatey/Website/Controllers/PackagesController.cs
```diff
@@ -145,20 +148,42 @@ public virtual ActionResult DisplayPackage(string id, string version, FormCollec
             // maintainers and reviewers cannot change the current status
             if (User.IsReviewer() || isMaintainer) status = package.Status;
 
+            if (status == PackageStatusType.Approved && !(
+                (package.PackageValidationResultStatus == PackageAutomatedReviewResultStatusType.Passing || package.PackageValidationResultStatus == PackageAutomatedReviewResultStatusType.Exempted)
+                    && (package.PackageTestResultStatus == PackageAutomatedReviewResultStatusType.Passing || package.PackageTestResultStatus == PackageAutomatedReviewResultStatusType.Exempted)
+                    && (package.PackageScanStatus == PackageScanStatusType.NotFlagged || package.PackageScanStatus == PackageScanStatusType.Exempted)
+                ))
```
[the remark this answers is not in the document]
gep13: :+1: this has also been addressed.

chocolatey/Website/Controllers/ApiController.cs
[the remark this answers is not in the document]
ferventcoder: Are you compiling that to .NET 4.x? Just curious.

chocolatey/Website/ViewModels/PackageViewModel.cs
```diff
@@ -95,16 +99,25 @@ public PackageViewModel(Package package)
...
         public string ExemptedFromVerificationReason { get; set; }
 
+        [Display(Name = "Scanner Exempted Reason")]
+        [StringLength(500)]
+        public string ExemptedFromScannerReason { get; set; }
+
+        [Display(Name = "Validator Exempted Reason")]
+        [StringLength(500)]
+        public string ExemptedFromValidatorReason { get; set; }
+
```
ferventcoder, line 120: 👍

chocolatey/Website/Services/PackageService.cs
```diff
@@ -1133,12 +1137,35 @@ public void ExemptPackageFromTesting(Package package, bool exemptPackage, string
...
+            if (exemptPackage)
+            {
+                if (!string.IsNullOrWhiteSpace(newComments)) newComments += "{0}".format_with(Environment.NewLine);
+                newComments += "Auto Verification Change - Verfication tests have been exempted.";
+            }
+            else
+            {
+                if (!string.IsNullOrWhiteSpace(newComments)) newComments += "{0}".format_with(Environment.NewLine);
+                newComments += "Auto Verification Change - Verfication tests have been set to unknown.";
+            }
```
ferventcoder, line 1150: Good to have this here. Thanks for adding

chocolatey/Website/Services/PackageService.cs
```diff
@@ -1133,12 +1137,35 @@ public void ExemptPackageFromTesting(Package package, bool exemptPackage, string
...
         }
 
-        public void ExemptPackageFromValidation(Package package)
+        public void ExemptPackageFromValidation(Package package, string reason, User reviewer)
         {
             package.PackageValidationResultStatus = PackageAutomatedReviewResultStatusType.Exempted;
             package.PackageValidationResultDate = DateTime.UtcNow;
+            package.ExemptedFromValidatorById = reviewer.Key;
+            package.ExemptedFromValidatorReason = reason;
```
ferventcoder, line 1158: Definitely good to be capturing this for auditing!

chocolatey/Website/Services/PackageService.cs
```diff
@@ -194,7 +194,7 @@ public PackageRegistration FindPackageRegistrationById(string id, bool useCache)
 
         public virtual Package FindPackageByIdAndVersion(string id, string version, bool allowPrerelease = true)
         {
-            return FindPackageByIdAndVersion(id, version, allowPrerelease, useCache: true);
+            return FindPackageByIdAndVersion(id, version, allowPrerelease, false);
```
ferventcoder, line 197: This change seems weird. What specifically was it about and does it get separated to some other issue (I haven't checked commits on where this came in as I'm in mid-review).

chocolatey/Website/Controllers/PackagesController.cs
```diff
@@ -191,12 +220,18 @@ public virtual ActionResult DisplayPackage(string id, string version, FormCollec
...
+                TempData["ErrorMessage"] = "There was a problem completing the operation.  Check validation summary for more information.";
+
                 return View("~/Views/Packages/DisplayPackage.cshtml", model);
             }
 
             if (isMaintainer && string.IsNullOrWhiteSpace(newComments))
             {
                 ModelState.AddModelError(String.Empty, "You need to provide comments.");
+
+                TempData["ErrorMessage"] = "There was a problem completing the operation.  Check validation summary for more information.";
```
ferventcoder, line 233: Oh goodness. We forgot that.... :/

ferventcoder: 👍

chocolatey/Website/Controllers/ApiController.cs
```diff
@@ -568,7 +583,17 @@ public virtual ActionResult SubmitPackageScanResults(string apiKey, string id, s
...
+
+            var message = "{0} has failed automated scanning.".format_with(package.PackageRegistration.Id);
+            if (packageScanStatus == PackageScanStatusType.NotFlagged)
+            {
+                message = "{0} has passed automated scanning.".format_with(package.PackageRegistration.Id);
+            }
+
+            packageSvc.UpdateSubmittedStatusAfterAutomatedReviews(package);
+
+            packageSvc.ChangePackageStatus(package, package.Status, package.ReviewComments, message, testReporterUser, testReporterUser, sendMaintainerEmail: true, submittedStatus: package.SubmittedStatus, assignReviewer: false);
```
ferventcoder, line 596: Little bit concerned about this here as it moved from minor change to ChangePackageStatus which could do something unintended for this section.

chocolatey/Website/Controllers/ApiController.cs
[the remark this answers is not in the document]
AdmiringWorm: Yes. I compiled and tested it while targeting .NET 4.0 Client Profile, 4.5, and 4.8 (not at the same time obviously 😕).

chocolatey/Website/Controllers/ApiController.cs
gep13: This is a necessary change, based on bringing the scanner fully into the moderation process.  Both package-validator and package-verifier call this method.

chocolatey/Website/Services/PackageService.cs
gep13: I made this change locally to ensure that I was always getting the latest package, since I was messing with changes directly in the database.  This wasn't meant to be included in the commit.  Thanks for catching this!

gep13:
@ferventcoder I believe I have addressed all the comments, and have added the last remaining parts to this PR.  Would be great if this could be reviewed again, and ideally taken for a spin.

chocolatey/Website/Infrastructure/Lucene/LuceneSearchService.cs
```diff
@@ -208,6 +208,7 @@ private static Package PackageFromDoc(Document doc)
                 DownloadCache = doc.Get("DownloadCache"),
                 PackageScanStatusForDatabase = doc.Get("PackageScanStatus"),
                 PackageScanResultDate = packageScanResultDate,
+                PackageScanFlagResultForDatabase = doc.Get("PackageScanFlagResult"),
```
ferventcoder, line 211: @gep13  PackageIndexEntity sets this I would say.  So you are going to need to set it in there. It's a Lucene Document that is it using to grab that from.

chocolatey/Website/Services/PackageService.cs
```diff
@@ -1021,31 +1023,33 @@ public void ChangeTrustedStatus(Package package, bool trustedPackage, User user)
...
-                //we don't do human moderation for prereleases
+                // we don't do human moderation for prereleases
                 if (package.IsPrerelease)
                 {
                     package.Listed = true;
                     package.Status = PackageStatusType.Exempted;
                 }
 
-                var trustedPackagePassingAutomatedReview = PackageIsTrusted(package) && passingVerification && passingValidation;
+                var trustedPackagePassingAutomatedReview = PackageIsTrusted(package) && passingScanner && passingVerification && passingValidation && scanFlagResult != PackageScanFlagResultType.Error;
```
ferventcoder, line 1052: 👍

chocolatey/Website/Services/PackageService.cs
```diff
@@ -1078,7 +1082,7 @@ public void ChangePackageTestStatus(Package package, bool success, string result
 
             packageRepo.CommitChanges();
 
-            var testComments = string.Format("{0} {1}.{2} This is not the only check that is performed so check the package page to ensure a 'Ready' status.{2} Please visit {3} for details.",
+            var testComments = string.Format("{0} {1}.{2} This is not the only check that is performed. Check the package page to ensure the package has reached 'Ready' status.{2} Please visit {3} for details.",
```
ferventcoder, line 1085: I'd almost love for this to report the status of checks that have passed `x/3` so the user knows exactly where they stand. This is where the most confusion for community maintainers is born out of (this message). 

They say it passed automated tests, but in reality it only passed one. So if the message could start out with something like "The package has passed 2/3 automated checks. It just passed x".

chocolatey/Website/Controllers/ApiController.cs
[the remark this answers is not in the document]
vexx32: Done! #941

chocolatey/Website/Views/Packages/DisplayPackage.cshtml
gep13: I have created a follow up issue here: https://github.com/chocolatey/chocolatey.org/issues/942 to capture this suggestion.

chocolatey/Website/Controllers/ApiController.cs
```diff
@@ -453,11 +453,11 @@ public virtual ActionResult SubmitPackageValidationResults(string apiKey, string
             package.PackageValidationResultDate = DateTime.UtcNow;
             package.PackageValidationResultStatus = PackageAutomatedReviewResultStatusType.Failing;
 
-            var message = "{0} has failed automated validation.".format_with(package.PackageRegistration.Id);
+            var message = "{0} has failed automated package validation.".format_with(package.PackageRegistration.Id);
             if (success)
             {
                 package.PackageValidationResultStatus = PackageAutomatedReviewResultStatusType.Passing;
-                message = "{0} has passed automated validation. It may have or may still fail other checks like testing (verification).".format_with(package.PackageRegistration.Id);
+                message = "{0} has passed automated package validation.".format_with(package.PackageRegistration.Id);
```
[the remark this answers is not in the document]
gep13: :+1:

chocolatey/Website/Controllers/ApiController.cs
```diff
@@ -634,7 +649,35 @@ public virtual ActionResult SubmitPackageScanResults(string apiKey, string id, s
...
+                message = "{0} has passed automated virus scanning.".format_with(package.PackageRegistration.Id);
+            }
+
+            if (packageScanFlagResult == PackageScanFlagResultType.Error)
+            {
+                // Send back to maintainer to address problem.
+                newSubmittedStatus = PackageSubmittedStatusType.Waiting;
+
+                if (!string.IsNullOrWhiteSpace(message)) message += "{0}".format_with(Environment.NewLine);
+                message += "At least one file within this package had greater than 10 detections.";
```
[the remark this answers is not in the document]
gep13: 👍

chocolatey/Website/Infrastructure/Lucene/LuceneSearchService.cs
gep13: Gotcha.  I thought I had caught all the places that this would be required, but looks like I missed this one.

chocolatey/Website/Services/PackageService.cs
gep13: I have made some changes to this, based on the suggestion that you made for when a package passes validation.  Given that there is a linear path through each stage of automated package testing now, I am not sure if we "need" the `x/3`.  A package either passes each stage, or it gets held in the queue.

gep13:
@ferventcoder @vexx32 I believe I have addressed all the comments that have been made as part of the review of this PR.  Please let me know if there are any other comments.  As requested, could either of you confirm whether you have had a chance to take this code for a spin on your environments?  It would be great to get some additional feedback on actually running the code, rather than just a review of the code.

vexx32 (commented):
@gep13 looks like there's still a couple minor display issues around the **Details** links and sometimes with reasons applied to each stage it goes a bit funny.

text bunched up with verification exemption

Looks like there should be line breaks before and after **Details** to match the others, but it renders the whole thing on a single line.

Also, when there is no exemption comment etc the **Details** link appears in a completely different location, off to the right side. The other **Details** links seem to appear on the line below, we should probably be consistent about that, one way or the other. There seems to be some inconsistencies in their placements...

two details links in very different locations

all bunched up

I also noticed one case where the `Exempt this version from X` failed to show in the checklist. If both the Validation and Verification checks are skipped/exempted on the same package, it seemed to prevent it from showing a checkbox to exempt from the scanner step. I was still able to advance the states by manually selecting `Exempted` and submitting a response as a moderator though.

Also, not sure of the purpose for the **Package Status** box having `Exempted` states anymore if those are going to be completely separate checkboxes (assuming they always show up when they might be needed, of course).

One last thing -- the first two automatic verification checks state `Passed/Failed/Exempt`, but the scanner uses `Successful/Failed/Exempt(ed?)`. We should probably have them all use the same terminology.

Nothing else of note seems to be breaking, which is lovely! I did hit a few cache issues and think for a bit I might have had some disappearing packages, but they all came back eventually so no issues there either.

gep13:
@designxsteph are you able to provide any assistance in the issues regarding placement of UI that @vexx32 mentions above?

designxsteph:
> @designxsteph are you able to provide any assistance in the issues regarding placement of UI that @vexx32 mentions above?

@gep13 @vexx32 I'm finding these display placement issues very strange because I know I have fixed these exact issues before. 🤔 I'll take a look at this today and figure out what's going on!

designxsteph:
@gep13 @vexx32 After my commits have been added back in, the display issues seem to be fixed. It would be worth double checking on your local as well to confirm.👍

vexx32:
Nice! Looks good on my end, thanks for that @designxsteph! 💖 

image

PR #919, opened 2020-08-03
(GH-613) Capture requestId when logging exception

gep13:
Adding a NewLine into the trace message results in it being "lost" when
filtering log based on requestId.  Write out exception information as
a separate trace line, so that it retains the requestId.

### DIFF
--- a/chocolatey/Website/Controllers/ApiController.cs
+++ b/chocolatey/Website/Controllers/ApiController.cs
@@ -281,7 +281,8 @@ any moderation related failures.",
                 }
                 catch (Exception ex)
                 {
-                    Trace.TraceError("[{0}] - Unable to delete temporary file at '{1}':{2} {3}.".format_with(requestId, temporaryFile, Environment.NewLine, ex.to_string()));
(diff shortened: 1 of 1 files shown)


chocolatey/Website/Controllers/ApiController.cs
```diff
@@ -281,7 +281,8 @@ private ActionResult CreatePackageInternal(string apiKey)
                 }
                 catch (Exception ex)
                 {
-                    Trace.TraceError("[{0}] - Unable to delete temporary file at '{1}':{2} {3}.".format_with(requestId, temporaryFile, Environment.NewLine, ex.to_string()));
+                    Trace.TraceError("[{0}] - Unable to delete temporary file at '{1}':{2} {3}.".format_with(requestId, temporaryFile));
+                    Trace.TraceError("[{0}]   - {1}", requestId, ex.to_string());
```
ferventcoder, line 285: You didn't remove the 2/3 items from the first, not sure it is necessary to separate this out though - mentioned it in the convo on this

PR #922, opened 2020-08-07
(GH-922) Show Content Inside Collapsed Text Editor

designxsteph:
There is a slight bug in the Easymde JS library used that when a text
editor is hidden on load inside a collapsed or tabbed item, the content
is not displayed until the editor is clicked on. A fix has been
put in place that will basically "refresh" the editor whenever the
hidden editor becomes visible. A new class must be applied to the
collapsible item of "text-editor-refresh" in order for this JS to
trigger.

More information on this bug and fix here:
https://github.com/Ionaru/easy-markdown-editor/issues/208#issuecomment-645656131

### DIFF
--- a/chocolatey/Website/Views/Packages/DisplayPackage.cshtml
+++ b/chocolatey/Website/Views/Packages/DisplayPackage.cshtml
@@ -1152,7 +1152,7 @@ salt '*' chocolatey.install @Model.Id.ToLower() version="@Model.Version" source=
                                         }
                                     </div>
                                     <button class="btn btn-sm btn-secondary mt-2" type="button" data-toggle="collapse" data-target="#test-exemptions" aria-expanded="false" aria-controls="test-exemptions">Show Testing Exemption Section</button>
-                                    <div class="collapse form-field" id="test-exemptions">
+                                    <div class="collapse form-field text-editor-refresh" id="test-exemptions">
                                         <ul class="mb-0 mt-3">
                                             <li>Exempting a package id from testing requires a reason.</li>
                                             @if (moderator)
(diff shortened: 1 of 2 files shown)


chocolatey/Website/Views/Packages/DisplayPackage.cshtml
```diff
@@ -1152,7 +1152,7 @@ salt '*' chocolatey.install @Model.Id.ToLower() version="@Model.Version" source=
                                         }
                                     </div>
                                     <button class="btn btn-sm btn-secondary mt-2" type="button" data-toggle="collapse" data-target="#test-exemptions" aria-expanded="false" aria-controls="test-exemptions">Show Testing Exemption Section</button>
-                                    <div class="collapse form-field" id="test-exemptions">
+                                    <div class="collapse form-field text-editor-refresh" id="test-exemptions">
```
designxsteph, line 1155: @gep13 Just a note on this. I know you added some more text editors [in this PR](https://github.com/chocolatey/chocolatey.org/pull/910). You will need to add the class `text-editor-refresh` to the parent collapse item of each editor as I did right here. Adding this class before this PR is excepted won't cause any bad side effects. Or maybe this is something we pull in after your PR, and I'll update this one to include the new instances once yours is accepted. Either way!

gep13: @designxsteph thanks for the heads up.  I am going to go ahead and get this PR merged in, so I will update my PR to include this new data attribute.

gep13:
@designxsteph thank you for figuring out what the issue was here, really appreciate it!

PR #925, opened 2020-08-12
(layout) Add Right Side Announcement Bar

designxsteph:
An announcement bar has been added that will appear on the right hand
side of the screen while on desktop, and full screen while on mobile.
The announcement bar is slightly different on the package listing page
then on the rest of the website, as it is static and always visible. On
the rest of the website, there is a new button that is located on the
right hand side of the screen that will show/hide this announcement
area when clicked. While on mobile, this switches to a button that is
added to the top navigation bar.

The announcements are set to "shuffle" on reload, so that a user will
potentially see different announcements while navigating throughout the
website.

### DIFF
--- a/chocolatey/Website/Content/scss/_base.scss
+++ b/chocolatey/Website/Content/scss/_base.scss
@@ -64,6 +64,10 @@ code:not([class*="language-"]) {
     text-decoration: underline;
 }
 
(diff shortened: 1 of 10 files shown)


designxsteph:
This is on test!

gep13:
@designxsteph thanks for getting this done, will get it released just now.

PR #928, opened 2020-08-13
(layout) Adjust Gitter Chat Box for Mobile

designxsteph:
While on mobile, the Gitter chat box is very tiny and unreadable. This
slight change in CSS adjust the chat box to take up the entire screen
on a mobile device so that it is easier to type and read.


gep13:
@designxsteph thanks for getting this updated.  I will get this pushed to chocolatey.org just now.

Issue #929, opened 2020-08-13
Many Outdated CSS and JS Libraries


designxsteph:
### What You Are Seeing?
There are a number of CSS and JS libraries that need to be updated. Some of these might have breaking changes and will require manual testing to be sure nothing has broken with the update.

- [x] Bootstrap
- [x] Font Awesome
- [x] jQuery
- [x] Anchor.js
- [x] EasyMDE
- [x] Clipboard.js
- [x] npm packages/ run npm audit fix
- [x] jquery-validation
- [x] knockout.js

designxsteph:
This is up on test. Everything should function the same as it has, so no notable change should be present just by navigating around the site.

PR #930, opened 2020-08-13
(GH-929) Update JS and CSS Libraries

designxsteph:
In this PR:

Updates a number of CSS and JS libraries that are used throughout the website. All npm vulnerabilities have also been fixed by running `npm audit fix`.

Libraries Updated:
Bootstrap, jQuery, Font Awesome, EasyMDE, Cliboard.js, Anchor.js, jquery-validation, knockout.js, npm packages


designxsteph:
This is up on test. Everything should function the same as it has, so no notable change should be present just by navigating around the site.

PR #932, opened 2020-08-15
(GH-613) Compute hash using stream 

gep13:
Rather than computing it from a byte[]. Using the ReadAllBytes method
is causing an OutOfMemory exception to occur with larger packages. The
ComputeHash method that is being used can also accept a Stream, rather
then a byte[], which should use less memory overall.

Relates to #613


gep13:
@ferventcoder @vexx32 I did a test against that current GenerateHash method, and verified that the hash that was generated from both the stream and byte[] methods was the same.

gep13:
@vexx32 thank you for the review.  Going to take this for a spin!

Issue #939, opened 2020-08-24
Announcement bar button not noticeable


designxsteph:
### What You Are Seeing?
Recently a new announcement bar has been added to the right side of the screen on most pages. Sometimes the button to expand this announcement bar is easy to miss, especially when it is on top of a blue header banner.

PR #940, opened 2020-08-24
(GH-939) Add Notification Badge to Notification Button

designxsteph:
Sometimes the button to the right side notification bar is not very
noticeable. To draw a bit more attention to this feature, a little red
badge has been added to the button if the user has not yet clicked it.
The badge also displays the number of "unread" notifications, much like
the notifications you might see on an app when it needs updating. Once
the user clicks on the button, this badge will disappear and a cookie
is set.

The next time the notification bar is updated, two hidden fields have
been added that will also need to change in order for the cookie to be
reset. Instructions on what to do have been added to the document.


designxsteph:
notification-desktop

notification-mobile

ferventcoder:
This is merged, but not yet deployed.

PR #946, opened 2020-08-25
(GH-944) Adjust Password Regex Special Characters

### LINKED ISSUE #944: Some special characters not included in password regex
### What You Are Seeing?
Some special characters are not included in the regex that forces a user to include a special character in their password. 
In this case, a user might use a special character and still receive an error.

### What is Expected?
I'd expect any special character to work.

Create an account and use a `-` as the special character.

designxsteph:
Adjusts the regex used to detect special characters by not including
any letter character, number, or white space instead of listing out
all the special characters.

Fixes #944


gep13 (approved):
LGTM!

gep13:
@designxsteph thanks for getting this fixed!

Issue #960, opened 2020-09-11
Prev/Next Buttons Do Not Work on Carousels


designxsteph:
### What You Are Seeing? 
The previous/next buttons are not working on any of the carousel items. When clicked, they do nothing, or freeze the carousel.

### What is Expected?
I would expect the buttons to work and move to the next slide.

### How Did You Get This To Happen? (Steps to Reproduce)
1. Go to the home page and click the prev/next buttons on the carousel item located under the webinar announcement (or any carousel on any page).

gep13:
@designxsteph this has been deployed, and is now live on chocolatey.org

PR #961, opened 2020-09-11
(GH-960) Make Carousel Prev/Next Buttons Function

designxsteph:
The prev/next buttons on any carousel item currently do not transition
between slides. This was due to a recent addition of the announcement
flyout, where a carousel was added with the same ID that may be found
on other pages. When the same ID is on the page twice, it breaks the
carousel. The ID has been changed to be unique and not match any other
carousel used throughout the website. Now, the prev/next buttons work
as intended.


gep13 (approved):
LGTM!

gep13:
@designxsteph this looks good to me!  Thanks for fixing this!

Issue #963, opened 2020-09-14
Convert Documentation to PDF


designxsteph:
### What You Are Seeing? 
It would be nice to be able to convert the documentation files into a PDF that can be printed off for later use.

### What is Expected?
A download PDF button to be present on each documentation page that when clicked, would download the page as PDF.

gep13:
I am going to go ahead and close this out, and I have got a new issue to track this here: https://github.com/chocolatey/docs/issues/85

PR #964, opened 2020-09-14
(GH-963) Convert Documentation .md Files to .pdf

designxsteph:
Converts all of the documentation files into a .pdf file by running the
gulp task `gulp convertDocsToPdf`. By doing so, a new folder named
`Temp_PDF` is created at the top level of the project, which all the
PDF's are then stored in. The intention is to run this gulp task,
upload the PDF's and then delete this temp folder.

A button has been added to each documentation page that allows the user
to download the current page they are viewing to a PDF document.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,10 +12,12 @@ App_Data
 /chocolatey/Website/Content/dist
 /chocolatey/Website/node_modules
 /chocolatey/Website/chocolatey-styleguide.zip
+/chocolatey/Website/Content/Pdf/chocolatey-pdf.min.css
 snags/
 _AzurePackage
 *.sln.docstates
 *.bak
+Temp_PDF
 
 .DS_Store
 Thumbs.db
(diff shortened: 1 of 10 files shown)


.gitignore
```diff
@@ -12,10 +12,12 @@ App_Data
 /chocolatey/Website/Content/dist
 /chocolatey/Website/node_modules
 /chocolatey/Website/chocolatey-styleguide.zip
+/chocolatey/Website/Content/Pdf/chocolatey-pdf.min.css
```
ferventcoder, line 15: Was this on purpose?

designxsteph: Yes, this file is compiled from scss when `gulp convertDocsToPdf` is ran, so we don't need to note any changes to this file. Any changes would be done and notated in the scss.

gep13:
@designxsteph I am going to go ahead and close out this PR, as the work for this would need to be done in the new docs repository.

I have created an issue to track this here: https://github.com/chocolatey/docs/issues/85

Issue #965, opened 2020-09-15
QDE v1 links need added to left side navigation in documentation area


designxsteph:
### What You Are Seeing? 
The QDE documentation was recently updated to v2, but there is no navigation links to the v1 docs in the left side navigation in the documentation area.

### What is Expected?
I would expect any version of the documentation to be available, incase a user has not upgraded and needs to find documentation relating to their current setup.

### How Did You Get This To Happen? (Steps to Reproduce)
Visit: https://chocolatey.org/docs/quick-deployment-setup-v1 and notice how the navigation on the left does not expand and nothing is highlighted.

ferventcoder:
This still needs pushed out.

PR #966, opened 2020-09-15
(GH-965) Add QDE v1 Links to Docs Nav

designxsteph:
Adds the QDE v1 links to the left side navigation in the documentation
area, under the existing QDE dropdown.

qdev1


steviecoaster:
Is there something we can do to make this a little less....repetitive? I don't know if we change the top level links or something, but I can see `QDE Setup` and `QDE Setup v1` causing a little confusion

gep13:
@designxsteph am I right in saying that the QDE v1 node would only be expanded should the user choose to expand it, or if they were to navigate directly to the page?  I think that would help alleviate any confusion.

gep13:
Looking into the future, having something like you get on the Microsoft Docs site would be pretty cool:

image

Where you get to a single page, and then you can select what version of the thing you want to look at.

pauby:
If this is functional and works, can we get it merged?

designxsteph:
@gep13 is correct, the v1 links will be collapsed inside the drop down unless the drop down is manually expanded or the user navigates to one of those pages. We can always merge this and then rethink a better naming convention if something else makes a bit more sense.

PR #971, opened 2020-09-22
(layout) Remove Puppet Webinar from Homepage

designxsteph:
Note: To be merged once the Puppet Webinar is over

Removes the Puppet webinar information from the homepage, as the event
date has passed and the replay is not available yet. This
replaces that information with the on-demand information for the
Deployments Webinar.


designxsteph:
homepage

PR #974, opened 2020-09-22
(GH-973) Prevent Form URL from Returning /policies

### LINKED ISSUE #973: Forms are returning the wrong URL on submission
### What You Are Seeing? 
If a user submits a form, the return URL after the submission always includes a `/policies/` in the URL.

### What is Expected?
The return URL to be the same or not include `/policies/` for no apparent reason.

1. Fill out any form
2. Once submitted, note the URL contains a `/policies/`

designxsteph:
(GH-973) Prevent Form URL from Returning /policies

Whenever a form was submitted, the URL being sent in and returned
always contained `/policies/` in the URL. With this update, this no
longer happens, and the forms are sent in from the URL in which they
come from, and also returned as such.

This bug was introduced in commit https://github.com/chocolatey/chocolatey.org/commit/9abd30456915145653706c7daf34aa2c9e1ba2a5 and happened whenever a form was references like:

`@using (Html.BeginForm("ContactSales", "Pages", FormMethod.Post, new { id = "i-recaptcha" }))`

Fixes #973

### DIFF
--- a/chocolatey/Website/App_Start/Routes.cs
+++ b/chocolatey/Website/App_Start/Routes.cs
@@ -386,7 +386,12 @@ namespace NuGetGallery
 
             routes.Add(new JsonRoute("json/{controller}"));
 
-            routes.MapRoute(RouteName.Policies, "policies/{action}", MVC.Pages.Terms());
+            routes.MapRouteSeo(
+               RouteName.Policies, "policies", new
+               {
+                   controller = "Pages",
+                   Action = "Terms"
+               });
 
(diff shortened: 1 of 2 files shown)


designxsteph:
@ferventcoder This is up on test.

This change corrects the problem we are seeing here, and still allows both /terms and /policies to work and show the same page. I'm unsure though of why this was actually causing this problem though with the form URLs.

ferventcoder (approved):
👍

Issue #972, opened 2020-09-22
Implement Chat Feature for Chocolatey for Business Questions


designxsteph:
It would be a nice enhancement to have some kind of chat feature for users that are interested in Chocolatey for business. This could connect to a member of the Chocolatey team that could answer common questions and help users find the solution they are looking for.

PR #976, opened 2020-09-23
(GH-972) Implement Chat Feature

designxsteph:
Implements a chat feature that users can ask/interact with Chocolatey
members about Chocolatey for Business. This is restricted to certain
pages that users might navigate to to find out more about the Business
features Chocolatey offers.

### DIFF
--- a/chocolatey/Website/Scripts/chat/chat.js
+++ b/chocolatey/Website/Scripts/chat/chat.js
@@ -1,0 +1,28 @@
+function chatBot(chatBotId) {
+    if (chatBotId) {
+        "use strict";
+
+        !function () {
+            var t = window.driftt = window.drift = window.driftt || [];
+            if (!t.init) {
+                if (t.invoked) return void (window.console && console.error && console.error("Drift snippet included twice."));
+                t.invoked = !0, t.methods = ["identify", "config", "track", "reset", "debug", "show", "ping", "page", "hide", "off", "on"],
+                    t.factory = function (e) {
+                        return function () {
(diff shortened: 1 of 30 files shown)


ferventcoder (changes requested):
See if we can move the main bulk of the function out to a js file, then load that in on bundles.reference. That way we can reduce the amount of duplication across multiple pages.

We can pass the chatbotid into the function once loaded and let it decide if it is null or not, and decide to load up. 

See if we can load it at the bottom as well, it doesn't seem like it needs the priority to load with top scripts.

designxsteph:
@ferventcoder This has been fixed and new version is up on test now.

chocolatey/Website/Scripts/chat/chat.js
```diff
@@ -0,0 +1,26 @@
+function chatBot(chatBotId) {
+    "use strict";
+
+    !function () {
```
gep13, line 4: @designxsteph should there still be a check in here to verify that the chatBotId isn't null, before attempting to use it?

gep13 (changes requested):

chocolatey/Website/Scripts/chat/chat.js
designxsteph: @gep13 Ah yeah, I added that in!

designxsteph:
@ferventcoder this should be on all the pages we want it on now, let me know if there are more. A new version of this is on test.

ferventcoder (approved):

Issue #993, opened 2020-10-16
install.ps1: Supporting Server 2012 R2 Core without WoW64 and making unzipping more robust


jberezanski:
Recently I wanted to use Chocolatey on a Windows Server 2012 R2 Core trimmed down so much that even the 32-bit subsystem (WoW64) was not present. This posed no problem for choco.exe or PowerShell, but there was the issue of actually installing Chocolatey. The existing install.ps1 uses 7-Zip (7za.exe) by default, which is a 32-bit application and won't work on such a system. Manually switching to the alternate method supported by install.ps1 would not help, because the system had PowerShell 4.0 (so no Expand-Archive cmdlet) and no GUI (so no unzipping by Explorer).

I was able to fix this quickly by hacking install.ps1 to use the ZipFile class present in .NET 4.5 and later, but it motivated me to take a long look at how unzipping is done in that script and I came up with a few improvements.

My thoughts were:
- all systems still supported by Microsoft today have a builtin mechanism for unzipping (and I do not mean Explorer);
- the install script should be as easy to use as possible, ideally not require the user to tweak anything;
- since some users are inconvenienced by an additional download, and it reduces the reliability of the install process, avoid doing that.

I changed the script so that it attempts these unzip methods in order, until one succeeds or all fail:
1) Expand-Archive - available with PowerShell 5+
2) System.IO.Compression.FileSystem - available with .NET 4.5+
3) 7-zip - requires additional download, does not work on pure 64-bit systems (Server Core with WoW64 subsystem removed)
4) Windows Explorer decompresion - does not work on Server Core

Method 1) should work on all modern systems (Win10, Server 2016+).
Method 2) covers older systems still supported by Microsoft (Win8.1, Server 2012 and 2012 R2).
Methods 3) and 4) remain for compatibility with even older, out-of-support OSes (such as Win7).

The user is able to control the list of methods by setting the `chocolateyUnzipMethods` environment variable to a list (separated by commas, semicolons, pipes or spaces) of method names, chosen out of this set:
PowerShell
DotNet
SevenZip
Explorer

For backward compatibility, the `chocolateyUseWindowsCompression` environment variable is still recognized, and if set to true and `chocolateyUnzipMethods` is not set, the method list becomes 'PowerShell,DotNet,Explorer'.

This change brings the following benefits:
- support for Server 2012/2012 R2 Core with 32-bit subsystem (WoW64) removed;
- no need for additional 7-Zip download on modern OSes;
[...]

PR #994, opened 2020-10-16
(GH-993) install.ps1: Make unzipping just-work

jberezanski:
Attempt these unzip methods in order, until one succeeds or all fail:
1) Expand-Archive - available with PowerShell 5+
2) System.IO.Compression.FileSystem - available with .NET 4.5+
3) 7-zip - requires additional download, does not work on pure 64-bit systems
           (Server Core with WoW64 subsystem removed)
4) Windows Explorer decompresion - does not work on Server Core

Method 1) should work on all modern systems (Win10, Server 2016+).
Method 2) covers older systems still supported by Microsoft (Win8.1,
Server 2012 and 2012 R2).
Methods 3) and 4) remain for compatibility with even older, out-of-support
OSes (such as Win7).

The user is able to control the list of methods by setting the
`chocolateyUnzipMethods` environment variable to a list (separated by
commas, semicolons, pipes or spaces) of method names, chosen out of this
set:
PowerShell
DotNet
SevenZip
Explorer

For backward compatibility, the `chocolateyUseWindowsCompression`
environment variable is still recognized, and if set to true and
`chocolateyUnzipMethods` is not set, the method list becomes
'PowerShell,DotNet,Explorer'.

This change brings the following benefits:
- support for Server Core with 32-bit subsystem (WoW64) removed
- no need for additional 7-Zip download on modern OSes
- automatic fallback, so admins should no longer need to manually set
  environment variables and retry

### DIFF
--- a/chocolatey/Website/Install.ps1
+++ b/chocolatey/Website/Install.ps1
@@ -25,7 +25,7 @@
 # For an explicit version of Chocolatey, please set $env:chocolateyVersion = 'versionnumber'
 # To target a different url for chocolatey.nupkg, please set $env:chocolateyDownloadUrl = 'full url to nupkg file'
 # NOTE: $env:chocolateyDownloadUrl does not work with $env:chocolateyVersion.
-# To use built-in compression instead of 7zip (requires additional download), please set $env:chocolateyUseWindowsCompression = 'true'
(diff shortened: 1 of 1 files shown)


ferventcoder:
@jberezanski there is already a pull request on the install.ps1 script. See https://github.com/chocolatey/chocolatey.org/pull/912. It's likely we are flipping to this. If you think there should be changes still after that is finished, I would ask that you resubmit a PR then?

PR #997, opened 2020-10-31
$null should be on the lefthand side of comparison

AlanFlorance:


ferventcoder:
See https://github.com/chocolatey/chocolatey.org/pull/912. We are about to pull that one in, and it is likely it's fixed these already. Let me know if this should remain open?

PR #1009, opened 2020-11-23
(events) Add 12 Days Event Information

designxsteph:
Adds a new event page for the upcoming holiday Chocolatey event. This
also adjust the home page and adds an event into the right sidebar
announcements.

The JS for the countdown clock is a work in progress and will be added
in a later commit.


ferventcoder (approved):
Looks fantastic!!

Issue #1019, opened 2020-12-04
Remove documentation pages from this repository


gep13:
The new Chocolatey Documentation site has a [new repository](https://github.com/chocolatey/docs), and all the files are now located there.  As such, the documentation files in this repository should be removed, so that there is no ambiguity about where the docs should be updated.

The corresponding [choco-wiki](https://github.com/chocolatey/choco-wiki) repository has already been archived.

PR #1022, opened 2020-12-04
(GH-1019) Remove the docs from this site

gep13:
These are now located in the new [docs](https://github.com/chocolatey/docs) repository.


ferventcoder (changes requested):
This will have some conflicts that will need to be looked at, but other than that, I'm good with merging it once you fix those up.

PR #1023, opened 2020-12-04
(GH-1021) Remove Doc Search Button and Docs JS

### LINKED ISSUE #1021: Documentation search doesn't work
Steps to reproduce

1. Go to chocolatey.org
1. Click the magnifying glass at the top right hand corner
1. Input a search term
1. Click the `Docs` button
1. You will get the following:

image

It has correctly redirected to the docs site, but there is nothing listening in that location anymore, since we are now using Algolia to do the searching.

I think it might be best to just remove the search for docs in this location.

@ferventcoder @designxsteph thoughts?

designxsteph:
In this PR:

* Removes the doc search button in the top navigation
* Removes doc specific JS that is no longer needed

Fixes #1021


gep13:
@designxsteph should the following be changed at the same time:

https://github.com/chocolatey/chocolatey.org/blob/master/chocolatey/Website/Views/Shared/_SearchBox.cshtml#L9-L11

designxsteph:
> @designxsteph should the following be changed at the same time:
> 
> https://github.com/chocolatey/chocolatey.org/blob/master/chocolatey/Website/Views/Shared/_SearchBox.cshtml#L9-L11

That line was removed because that is what was redirecting the search over to the docs. When removing it, it now returns the `/search` url if the user is on any page other than the packages page. In that case it returns the `/packages` url. The pages are returning the same view, just different URL. This is how it has been for a while now, but not really sure that makes much sense/difference.

gep13:
Ignore my previous comment.  For some reason I missed that you had modified that file as part of this PR.  I was expecting that this file needed to change, and I couldn't see it for some reason.  That changes that you have made are exactly what is required.

Ignore me!

PR #1040, opened 2021-01-06
Upgrade to .NET Framework 4.5.2

### LINKED ISSUE #1039: Update Website project to use .NET Framework 4.5.2
All Package Services:

* Package Cleaner
* Package Validator
* Package Verifier
* Package Scanner
* Package Cacher

Have been updated to use .NET Framework 4.5.2, so it is time to also update the website to use the same base framework.

gep13:
In order to accomplish this, it was necessary to upgrade EntityFramework
package to 5.0.0.  This was due to the fact that some attributes that
are being used on the entity classes have moved to a new namespace.
The necessary attribute changes have been made.

In addition, it was necessary to add the NotMapped attribute to a number
of properties, otherwise EF was changing that a migration was required
to be added.  My assumption here is that in EF 5.0.0 there has been a
change in how mapped properties were discovered. In EF 4.3.1 the
NotMapped attribute was not required, but now it is.

Where necessary, NuGet packages were uninstalled and re-installed in
order to re-target to .NET Framework 4.5.2.

Fixes #1039 

I am going to put this up on the Test Site, and I will comment on this PR when this is live.

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -164,7 +164,6 @@ Task("Build")
         PlatformTarget = PlatformTarget.MSIL,
         Targets = { "Build" }
     }
-    .WithProperty("CodeAnalysis", "true")
     .WithProperty("OutDir", buildData.BuildArtifactsDirectoryPath.FullPath)
     );
 });
--- a/chocolatey/Website/Entities/Course.cs
+++ b/chocolatey/Website/Entities/Course.cs
@@ -1,15 +1,15 @@
-﻿// Copyright 2011 - Present RealDimensions Software, LLC, the original 
+﻿// Copyright 2011 - Present RealDimensions Software, LLC, the original
 // authors/contributors from ChocolateyGallery
 // at https://github.com/chocolatey/chocolatey.org,
-// and the authors/contributors of NuGetGallery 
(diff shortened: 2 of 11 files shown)


gep13:
@ferventcoder I ended up staying at EntityFramework 5.0.0 rather than moving to 6.x.  In order to get to EF 6.x, we need to remove usage of the MiniProfiler.EF package, and instead use MiniProfiler.EF6, but this requires .NET Framework 4.6.1.  I figure that this could be done as a future enhancement.

gep13:
@ferventcoder this is now live on the test site.

ferventcoder (approved):
As far as changes, this looks great. Still need to dig into how it works up on the live site though!

ferventcoder:
Also, this is awesome!

gep13:
@ferventcoder been playing with this a little bit today, and don't think this is ready to be merged yet.  Seeing some issues with pushing packages.

gep13:
@ferventcoder this should now be ready for a review/test.  I have tested the publishing of packages locally in my test environment, and it is now working, where it was causing an exception for me yesterday.

